PR #11, opened 2025-08-19
Render markdown in terminal using marked-terminal

billycao:
before:
CleanShot 2025-08-16 at 21 16 57@2x

after:
CleanShot 2025-08-18 at 17 18 53@2x

CleanShot 2025-08-18 at 17 44 07@2x


reissbaker (changes requested):
I like rendering Markdown — but `marked-terminal` seems semi-abandonware (no updates in 7 months), and it looks like it requires locking `marked` to a version that's four major versions behind current (v12 instead of v16).

It also doesn't seem like a massive project to reimplement just the `marked-terminal` piece on top of latest `marked` v16, since the original codebase is less than 1k LOC and a lot of `marked-terminal` is reimplementing whitespace reflows (which Ink already handles for us), along with various extra features we don't really need (e.g. turning `:smile:` into a :smile: emoji, rendering JS using syntax highlighting but no other language, etc); reimplementing classes (!!) using the ES3-era `prototype` syntax, etc.

IMO it would be worth doing the port; it might even be a decent testbed for Octo itself, ie see if Octo can just do the manual, minimal TypeScript port using latest `marked`.

billycao:
Closing in favor of reimplementing from scratch :)

PR #12, opened 2025-08-19
Allow pressing up and down in input box for input history

billycao:
https://github.com/user-attachments/assets/c5c64136-5376-45da-a033-cb295cb0bd64

This was written almost entirely by LLM, I just reviewed and cleaned up the configs. :)

Reuses the `.local/share/octofriend` state directory to persist between sessions like Claude Code does.

### DIFF
--- a/source/components/input-with-history.tsx
+++ b/source/components/input-with-history.tsx
@@ -13,14 +13,12 @@ interface Props {
 
 export const InputWithHistory = React.memo((props: Props) => {
   const themeColor = useColor();
-  const [isNavigating, setIsNavigating] = useState(false);
   const [currentIndex, setCurrentIndex] = useState(-1);
   const [originalInput, setOriginalInput] = useState("");
 
   useInput((input, key) => {
     if (key.upArrow) {
-      if (!isNavigating) {
-        setIsNavigating(true);
+      if (currentIndex === -1) {
         setOriginalInput(props.value);
       }
 
@@ -35,7 +33,7 @@ export const InputWithHistory = React.memo((props: Props) => {
 
     if (key.downArrow) {
       const history = props.inputHistory.getCurrentHistory();
-      if (!isNavigating || history.length === 0) return;
+      if (currentIndex === -1 || history.length === 0) return;
 
       if (currentIndex < history.length - 1) {
         const newIndex = currentIndex + 1;
@@ -44,7 +42,6 @@ export const InputWithHistory = React.memo((props: Props) => {
       } else {
         // Reset to original input
         setCurrentIndex(-1);
-        setIsNavigating(false);
         props.onChange(originalInput);
       }
       return;
@@ -52,8 +49,7 @@ export const InputWithHistory = React.memo((props: Props) => {
 
     // Reset navigation state when user types anything else
     if (input || key.return || key.escape || key.backspace || key.delete) {
-      if (isNavigating) {
-        setIsNavigating(false);
+      if (currentIndex !== -1) {
         setCurrentIndex(-1);
         setOriginalInput("");
       }
@@ -65,15 +61,13 @@ export const InputWithHistory = React.memo((props: Props) => {
       props.inputHistory.appendToInputHistory(props.value.trim());
     }
 
-    setIsNavigating(false);
     setCurrentIndex(-1);
     setOriginalInput("");
     props.onSubmit();
   };
 
   const handleChange = (value: string) => {
-    if (isNavigating) {
-      setIsNavigating(false);
+    if (currentIndex !== -1) {
       setCurrentIndex(-1);
       setOriginalInput("");
     }
--- a/source/input-history/index.ts
+++ b/source/input-history/index.ts
@@ -1,4 +1,4 @@
-import { sql } from "drizzle-orm";
+import { sql, notInArray, desc } from "drizzle-orm";
 import { db } from "../db/db.ts";
 import { inputHistoryTable } from "./schema/input-history-table.ts";
 
@@ -6,19 +6,14 @@ const MAX_HISTORY_ITEMS = 100;
 const MAX_HISTORY_TRUNCATION_BATCH = 20;
 
 export async function loadInputHistory(): Promise<InputHistory> {
-  try {
-    const historyRecords = await db().query.inputHistoryTable.findMany({
-      orderBy: (table, { asc }) => asc(table.id),
-      limit: MAX_HISTORY_ITEMS,
-    });
+  const historyRecords = await db().query.inputHistoryTable.findMany({
+    orderBy: (table, { asc }) => asc(table.id),
+    limit: MAX_HISTORY_ITEMS,
+  });
 
-    const history = historyRecords.map(({ input }) => input);
+  const history = historyRecords.map(({ input }) => input);
 
-    return new InputHistory(history);
-  } catch (error) {
-    console.warn("Failed to load input history:", error);
-    return new InputHistory([]);
-  }
+  return new InputHistory(history);
 }
 
 export class InputHistory {
@@ -34,28 +29,23 @@ export class InputHistory {
 
     this.history.push(input);
 
-    try {
-      await db()
-        .insert(inputHistoryTable)
-        .values({ input });
+    await db().insert(inputHistoryTable).values({ input });
 
-      await this.truncateOldEntries();
-    } catch (error) {
-      console.warn("Failed to save input history:", error);
-    }
+    await this.truncateOldEntries();
   }
 
   private async truncateOldEntries(): Promise<void> {
-    if (this.history.length % MAX_HISTORY_TRUNCATION_BATCH === 0) {
-      db().run(sql`
-        DELETE FROM ${inputHistoryTable}
-        WHERE id NOT IN (
-          SELECT id FROM ${inputHistoryTable}
-          ORDER BY id DESC
-          LIMIT ${MAX_HISTORY_ITEMS}
-        )
-      `);
-    }
+    db().transaction((db) => {
+      const historyToKeep = db
+        .select({ id: inputHistoryTable.id })
+        .from(inputHistoryTable)
+        .orderBy(desc(inputHistoryTable.id))
+        .limit(MAX_HISTORY_ITEMS);
+
+      db.delete(inputHistoryTable)
+        .where(notInArray(inputHistoryTable.id, historyToKeep))
+        .run();
+    });
   }
 }
 


reissbaker (changes requested):
This is a nice feature! But, I think there's a bug: if you're running two Octofriend instances (some people run multiple agents at a time to work on multiple things at once), the two Octos will clobber each other's history, since the lists are stored in-memory and neither process knows about the other.

I think from an architectural perspective, when you need to load state async upfront, it's also nice to do that in the `cli.tsx` prior to rendering the main App, so that the app doesn't need to deal with async loading states and possibly-null data while the initial async call loads (e.g. this is why we load the config and metadata upfront). This feels similar: we should probably load it upfront.

As a last note: from a UX perspective, it seems like it'd be nice to segment the history by working dir — i.e. allow navigating the history *for this project* rather than navigating globally.

I think the *simplest* possible implementation is probably a JSONL file, so that you can just do file appends and not clobber other Octo instances. But, you'll still have some wonky write contention issues to deal with: specifically, pruning the old entries so the file doesn't grow infinitely large, without the two Octos overwriting each other's history. It's technically possible with some cleverness, but...

In general I think there's value in just biting the bullet and using `sqlite`: we can store the DB file in `~/.local/share/octofriend/`, and it'll handle concurrency for us. There's definitely more persistence features that would be nice to build and will be a lot less painful via SQLite.

I'll do a minimal port of the existing `metadata.json5` file to SQLite, and then we can rebuild this more simply on top of that!

reissbaker:
FYI, the SQLite+Drizzle setup and update notification port is done, so feel free to build this on top of SQLite!

billycao (commented):
Ty for review! Removed the json impl in favor of the sqlite impl :). PTAL.

History is read once and written once in cli.tsx.

Also added in-memory sqlite for test environments.

drizzle.config.ts
```diff
@@ -1,7 +1,8 @@
 import type { Config } from "drizzle-kit";
 
 export default {
-  schema: "./source/**/schema/*.ts",
```
reissbaker, line 4: I think I prefer the original version here — the original meaning of this schema definition was to find all files in special `schema/` directories, and consider them as the schema; this also enforced that schemas were placed in those special subdirectories. This new version *works* (it loads the schemas, since they're all imported in `schema.ts` anyway), but it doesn't enforce consistent directory schemes for where the individual DB schema files live.

I think you made this change because you placed your schema in `source/input/schema.ts`, and it didn't pick up the change. If you move your schema definition to: `source/input/schema/input-history-table.ts`, it'll work with the original way of doing things, and enforce consistent directory structure.

source/input-history/index.test.ts
```diff
@@ -0,0 +1,68 @@
+import { describe, it, expect, beforeAll, beforeEach } from 'vitest';
+import { db } from '../db/db.ts';
+import { migrate } from '../db/migrate.ts';
+import { inputHistoryTable } from './schema.ts';
+import { getCurrentHistory, appendToInputHistory, loadInputHistory, saveInputHistory, _exportedForTest } from './index.ts';
+
+describe('Input History', () => {
+  beforeAll(async () => {
+    await migrate();
```
reissbaker: Nit: I think this should probably be placed in a single file referenced in the Vitest `setupFiles` config: https://vitest.dev/config/#setupfiles

That way, test authors only need to write this boilerplate once, instead of each file needing it at the beginning.

source/input-history/index.ts
```diff
@@ -0,0 +1,91 @@
+import { asc, count, inArray } from "drizzle-orm";
+import { db } from "../db/db.ts";
+import { inputHistoryTable } from "./schema.ts";
+import { expectOne } from "../db/query.ts";
+
+const MAX_HISTORY_ITEMS = 100;
+
+let currentHistory: { input: string; timestamp: Date }[] = [];
+let isInitialized = false;
```
reissbaker: It seems easy to forget to initialize this. I think we can use the type system here to make illegal states unrepresentable; one way to do it would be to have a class like:

```typescript
class InputHistory {
  constructor(private readonly history: Array<{ input: string, timestamp: Date }>) {}
  getInputHistory() { return [ ...this.history ] }
  async append(input: string) {
    if(input.trim()) {
      // Inser the history into the in-memory array, and add it to the table
    }
  }
}
```

Then make sure that `<App>` takes an instance of `InputHistory` as a prop, and you're guaranteed callers will be forced to instantiate this prior to running an App: they need to pass an instance into App, and the instance demands that they pass in the history array (so they'll need to load it ahead of time).

source/input-history/index.ts
```diff
@@ -0,0 +1,91 @@
...
+      input: record.input,
+      timestamp: record.timestamp
+    }));
+    isInitialized = true;
+  } catch (error) {
+    console.warn("Failed to load input history:", error);
+  }
+}
+
+export async function saveInputHistory(): Promise<void> {
```
reissbaker: I think you should probably just append to the table immediately in the same function where you update the in-memory history; that way, you can't forget to update the table, and since it's just a single-row insert it's super fast.

If you do that, it also slightly simplifies the table: you don't need to have an index on timestamps — you can just `ORDER BY id` and get insertion-order input.

source/input-history/index.ts
```diff
@@ -0,0 +1,91 @@
...
+        .select({ id: inputHistoryTable.id })
+        .from(inputHistoryTable)
+        .orderBy(asc(inputHistoryTable.timestamp))
+        .limit(totalCount - MAX_HISTORY_ITEMS);
+
+      if (recordsToDelete.length > 0) {
+        const idsToDelete = recordsToDelete.map(r => r.id);
+        await db()
+          .delete(inputHistoryTable)
+          .where(inArray(inputHistoryTable.id, idsToDelete));
```
reissbaker: Nit, but it's slightly simpler and faster to do this with a CTE or subquery, rather than doing two queries: you don't have to think about whether there's race conditions between the time you call the select query and the time you execute the second delete query, since it's all guaranteed to be atomic.

reissbaker (changes requested):
This is much improved! Left some comments on implementation. I might take this over while you're at Burning Man like we chatted about :smile: Since I think it's pretty close, and it's a nice feature.

drizzle.config.ts
billycao: Done. :)

source/input-history/index.test.ts
billycao: Done. :)

source/input-history/index.ts
billycao: Good point, done :) Also cleaned up the truncation script and added a unit test for truncation.

source/input-history/index.ts
billycao: Done. :) Also updated unit tests.

source/input-history/index.ts
billycao: Ended up rewriting this in a single query w/ subquery before seeing this comment, so good feedback :P Done. :)

billycao (commented):
Ty for review, PTAL :)

Rebased on main and simplified the table and impl per your feedback and smoke tested again locally.

source/input-history/index.ts
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  private async truncateOldEntries(): Promise<void> {
+    if (this.history.length % MAX_HISTORY_TRUNCATION_BATCH === 0) {
+      db().run(sql`
+        DELETE FROM ${inputHistoryTable}
+        WHERE id NOT IN (
+          SELECT id FROM ${inputHistoryTable}
+          ORDER BY id DESC
+          LIMIT ${MAX_HISTORY_ITEMS}
```
reissbaker: Handwritten SQL is sometimes necessary for advanced features, but — doesn't Drizzle support subqueries natively? Generally I strongly prefer using the Drizzle queries so that if the schema changes it's impossible to forget to update the query.

source/input-history/index.ts
```diff
@@ -0,0 +1,62 @@
...
+        .values({ input });
+
+      await this.truncateOldEntries();
+    } catch (error) {
+      console.warn("Failed to save input history:", error);
+    }
+  }
+
+  private async truncateOldEntries(): Promise<void> {
+    if (this.history.length % MAX_HISTORY_TRUNCATION_BATCH === 0) {
```
reissbaker: You should always run this query: you don't know if there's another Octo process running, so you can't trust this process's in-memory representation of the history length. SQlite should be fast — much faster than the LLM call — so it's fine to run this on every new entry!

source/input-history/index.ts
```diff
@@ -0,0 +1,62 @@
+import { sql } from "drizzle-orm";
+import { db } from "../db/db.ts";
+import { inputHistoryTable } from "./schema/input-history-table.ts";
+
+const MAX_HISTORY_ITEMS = 100;
+const MAX_HISTORY_TRUNCATION_BATCH = 20;
+
+export async function loadInputHistory(): Promise<InputHistory> {
+  try {
```
reissbaker: Generally speaking I'm against `try/catch` unless you expect this to fail for some knowable reason. We shouldn't expect that reading or writing to a local SQLite will fail, so we shouldn't add `try/catch` blocks.

source/input-history/index.ts
```diff
@@ -0,0 +1,62 @@
...
+    return this.history;
+  }
+
+  async appendToInputHistory(input: string): Promise<void> {
+    const trimmedInput = input.trim();
+    if (!trimmedInput) return;
+
+    this.history.push(input);
+
+    try {
```
reissbaker: ditto re: this shouldn't fail so we shouldn't wrap it in `try`/`catch`

source/components/input-with-history.tsx
```diff
@@ -0,0 +1,93 @@
...
+interface Props {
+  inputHistory: InputHistory,
+  value: string;
+  onChange: (s: string) => any;
+  onSubmit: () => any;
+}
+
+export const InputWithHistory = React.memo((props: Props) => {
+  const themeColor = useColor();
+  const [isNavigating, setIsNavigating] = useState(false);
```
reissbaker: I think you can remove this state and replace it with:

```typescript
const isNavigating = currentIndex === -1;
```

That way you can't forget to keep the two states in sync, since they track the same thing.

source/input-history/index.ts
billycao: Whoops, good catch. Done.

source/input-history/index.ts
billycao: Good point, done.

source/input-history/index.ts
billycao: Fair, done.

source/input-history/index.ts
billycao: Done

source/components/input-with-history.tsx
billycao: Good call, done.

billycao (commented):
Ty for the review. PTAL :)

PR #13, opened 2025-08-19
Pressing ctrl+c in input once clears text, twice exits app

billycao:
CleanShot 2025-08-18 at 18 16 21@2x

Feel like terminal users (like myself) are used to ctlr+c to clear input. :P

Kept most of the handling code in `components/ctrl-c-provider.tsx` to keep `app.tsx` clean. Also sets `exitOnCtrlC: false` to disable Ink's default handling of `SIGINT` so we can call `exit()` manually.

### DIFF
--- a/source/components/ctrl-c-provider.tsx
+++ b/source/components/ctrl-c-provider.tsx
@@ -6,25 +6,20 @@ type CtrlCContextType = {
   registerClearInputFn: (clearFn: () => void) => void;
 };
 
-const CtrlCContext = createContext<CtrlCContextType>({
-  ctrlCPressed: false,
-  registerClearInputFn: () => {},
-});
-
-export function useCtrlC() {
-  return useContext(CtrlCContext);
-}
-
 type CtrlCProviderProps = {
   children: React.ReactNode;
 };
+
+export const CtrlCContext = createContext<CtrlCContextType>({
+  ctrlCPressed: false,
+  registerClearInputFn: () => {},
+});
 
 export function CtrlCProvider({ children }: CtrlCProviderProps) {
   const [ctrlCPressed, setCtrlCPressed] = useState(false);
   const clearInputRef = useRef<(() => void) | null>(null);
   const { exit } = useApp();
 
-  // Global input handler to capture Ctrl+C
   useInput((input, key) => {
     if (key.ctrl && input === 'c') {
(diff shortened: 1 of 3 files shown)


source/components/ctrl-c-provider.tsx
```diff
@@ -0,0 +1,49 @@
...
+          clearInputRef.current();
+        }
+        setCtrlCPressed(true);
+        setTimeout(() => setCtrlCPressed(false), 2000);
+      }
+    }
+  });
+
+  const registerClearInputFn = (clearFn: () => void) => {
+    clearInputRef.current = clearFn;
```
reissbaker: I think this is a memory leak — `clearFn` will persist as long as `CtrlCProvider` does, and `CtrlCProvider` persists forever since it wraps the root nodes of the app.

It's true that the current use doesn't leak memory, since there's only one thing that ever calls it (`BottomBar`) and it overwrites itself on every re-render. But I think we probably shouldn't expose this: it feels pretty footgun-y.

An alternative that I think would feel similarly nice from a DX perspective but wouldn't leak memory IMO is to expose a function like:

```typescript
export function useCtrlC(callback: () => void) {
  useInput((input, key) => {
    if(key.ctrl && input === "c") callback();
  });
}
```

Then `CtrlCProvider` could use `useCtrlC(...)`, and `BottomBar` could as well. Since `CtrlCProvider` auto-exits the app on second ctrl-c, BottomBar doesn't need to be aware of anything about ctrl-c: either it'll run, or the app will exit.

For example, in `CtrlCProvider`:

```typescript
useCtrlC(() => {
  if(ctrlCPressed) exit();
  else {
    setCtrlCPressed(true);
    setTimeout(() => setCtrlCPressed(false), 2000);
  }
})
```

And in `BottomBar`, rather than using `useEffect` and registering stuff, you could:

```typescript
useCtrlC(() => {
  setQuery("");
});
```

In fact, with this implementation, I don't think you need React Contexts at all: `CtrlCProvider` can just be an ordinary component (`ExitOnDoubleCtrlC`?), since it doesn't need to expose anything to child nodes: it just exits after two ctrl-c keypresses in quick succession.

billycao: Excellent points, much cleaner and simpler impl. Ty :)

PR #17, opened 2025-08-19
Enable ctrl-a, ctrl-e, and ctrl-w shortcuts in user input box

billycao:
Re-implements the `TextInput` component to enable terminal like shortcuts.

Also tested with the other inputs, like the onboarding name and custom model and inputs.

https://github.com/user-attachments/assets/ae2a01c1-a9f3-485a-bf26-2ae16dec6e44

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
 		"ink": "^5.2.0",
 		"ink-select-input": "^6.0.0",
 		"ink-spinner": "^5.0.0",
-		"ink-text-input": "^6.0.0",
 		"json-schema": "^0.4.0",
 		"json5": "^2.2.3",
 		"openai": "^5.9.0",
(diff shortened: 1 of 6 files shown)


reissbaker (changes requested):
I like allowing the Emacs-style keyboard shortcuts, but I think it'd be worth just copying the full [`ink-text-input` implementation](https://github.com/vadimdemedes/ink-text-input/blob/master/source/index.tsx) and modifying the source; there are some nice features we should probably be using (i.e. masking input for API key input) and it'll prob be a little annoying to copy those over piecemeal in separate PRs.

billycao:
Ty for review, PTAL :)

While I was here I added some other common emacs style hotkeys that I noticed were also supported in Claude Code.

For ease of review the 2nd to last commit is just the ink-text-input impl with some style changes, and the last commit contains the hotkey additions. They were mostly implemented by LLM based on my existing impl, and I smoke tested they all seem to do the right thing.

reissbaker (approved):
love it! merging

PR #19, opened 2025-08-20
Custom implementation of markdown terminal rendering with marked

billycao:
As described. Spent a bunch of time cleaning up corner cases and little bugs.

Looks pretty decent in my color scheme. Also checked that it looked sane and decent in a couple other schemes in iterm2.

Uses `string-width` to get visual width of characters for table column rendering.

Interesting enough, in support of octofriend, this was one of the few cases where octo was definitively better than Claude Code from scratch. However it still struggled with subsequent direction (high token costs, long output delays, subpar output), at least with GLM-4.5.

My methodology ended up generating the initial code with octo/GLM-4.5 and using Claude Code to fine tune the little UI nits and improve aesthetics.

Demo script (also tested with some queries in octo itself):

CleanShot 2025-08-19 at 23 11 18@2x
CleanShot 2025-08-19 at 23 11 25@2x
CleanShot 2025-08-19 at 23 11 30@2x
CleanShot 2025-08-19 at 23 11 37@2x

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,8 +29,10 @@
 				"ink-text-input": "^6.0.0",
 				"json-schema": "^0.4.0",
 				"json5": "^2.2.3",
+				"marked": "^16.2.0",
 				"openai": "^5.9.0",
 				"react": "^18.2.0",
+				"string-width": "^7.2.0",
 				"structural": "^0.0.29",
 				"zustand": "^5.0.3"
 			},
@@ -2796,6 +2798,18 @@
 			"license": "MIT",
 			"dependencies": {
 				"@jridgewell/sourcemap-codec": "^1.5.0"
+			}
+		},
+		"node_modules/marked": {
+			"version": "16.2.0",
+			"resolved": "https://registry.npmjs.org/marked/-/marked-16.2.0.tgz",
+			"integrity": "sha512-LbbTuye+0dWRz2TS9KJ7wsnD4KAtpj0MVkWc90XvBa6AslXsT0hTBVH5k32pcSyHH1fst9XEFJunXHktVy0zlg==",
+			"license": "MIT",
+			"bin": {
+				"marked": "bin/marked.js"
+			},
+			"engines": {
+				"node": ">= 20"
 			}
 		},
 		"node_modules/math-intrinsics": {
--- a/package.json
+++ b/package.json
@@ -22,10 +22,10 @@
 	"files": [
 		"dist",
 		"source",
-    "drizzle",
-    "drizzle.config.ts",
-    "CHANGELOG.md",
-    "IN-APP-UPDATES.txt",
+		"drizzle",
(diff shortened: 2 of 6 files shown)


billycao:
Done, PTAL. :)

Had LLM update the code and manually cleaned up the rendering and some code nits. You can checkout and `npx tsx scripts/dev/markdown-demo.ts` for a quick demo or ofc ask octo.

billycao:
Inspired by a user, added syntax highlighting for code in markdown based on `highlight.js`. Also updated `npx tsx scripts/dev/markdown-demo.ts` to be more comprehensive.

source/markdown/highlight-code.tsx
```diff
@@ -0,0 +1,159 @@
...
+      });
+    }
+
+    currentIndex = closingTagStart + closingTag.length;
+  }
+
+  return segments;
+}
+
+function renderCodeLine(segments: CodeSegment[]): React.ReactElement {
```
reissbaker: Nit: this is a React component, it should be named with capital CamelCase (maybe `function CodeLine`?) and instantiated with `<CodeLine ...>` rather than being called as a function

source/markdown/highlight-code.tsx
```diff
@@ -0,0 +1,159 @@
+import { Text } from "ink";
+import hljs from "highlight.js";
+import React from "react";
+
+export function highlightCode(code: string, language?: string): React.ReactElement[] {
```
reissbaker: Nit: this is a React component as well

source/markdown/index.tsx
```diff
@@ -0,0 +1,372 @@
+import React from "react";
+import { Box, Text } from "ink";
+import { marked, MarkedToken, Token, Tokens } from "marked";
+import stringWidth from "string-width";
+import { isImageToken, isLinkToken, isTextToken, isStrongToken, isEmToken, isDelToken, isCodespanToken } from "./types.ts";
+import { highlightCode } from "./highlight-code.tsx";
+
+export function renderMarkdown(markdown: string): React.ReactElement {
```
reissbaker: Nit: this is also a React component

reissbaker (commented):
Generally looks good! Left a few small nits — I *think* calling things as React components rather than functions inside `{ ... }` can help with tree-diffing performance :)

source/markdown/highlight-code.tsx
billycao: Done! :)

source/markdown/highlight-code.tsx
billycao: This is actually a fn that returns an array of components right? :P But agree that that's silly and restructured this to be a component instead.

source/markdown/index.tsx
billycao: Done! :)

PR #24, opened 2025-08-29
feat: allow both 'args' and 'arguments' in MCP server config

italoalmeida0:
- Add support for both 'args' and 'arguments' fields in McpServerConfig
- Add validation to prevent using both fields simultaneously
- Update MCP tool handling to use either format
- Add .idea to .gitignore


reissbaker:
Hm. I think it's better to have a single config format and validate it, rather than accepting two keys that do the same thing.

PR #31, opened 2025-09-14
MCP v2: Core infrastructure security and environment variables

dayofdoom:
This focused refactor introduces essential MCP security fixes and environment variable support without any Kagi-specific code or examples.

### Configuration Schema (source/config.ts)
- Added `env: t.optional(t.dict(t.str))` to McpServerConfigSchema
- Enables environment variables for any MCP server configuration

### MCP Tool Security (source/tools/tool-defs/mcp.ts)
- **Input Sanitization**: Added sanitizeText(), sanitizeName(), sanitizeMcpArgs()
- **Command Injection Protection**: Validate and sanitize all subprocess commands/args
- **Cache Integrity**: Configuration-aware client caching with change detection
- **Memory Safety**: Bounds checking with safeNumber() prevents overflow
- **Error Handling**: Sanitized error messages prevent information leakage

### System Prompt Security (source/system-prompt.ts)
- Added try-catch for robust MCP server connection handling
- Sanitize server names and tool descriptions for display
- Skip servers that fail to connect gracefully

### Environment Variables Support
- Configurable environment variables per MCP server
- Merged with process environment variables
- Sanitized and validated to prevent injection
- Backwards compatible with existing configurations

### Security Improvements
- Control character removal from all inputs
- Safe character filtering for names and paths
- Argument limits to prevent abuse
- Type safety validation throughout

### Cache Management
- Hash-based configuration change detection
- Automatic client cleanup on config changes
- Memory-efficient connection management

## Minimal Scope

This implementation focuses exclusively on:
- Core MCP infrastructure
- Generic security improvements
- Environment variable functionality
- No Kagi-specific code or examples
- No vendor-specific documentation

The changes are backwards compatible and apply security best practices to the MCP infrastructure while keeping the scope extremely focused.


reissbaker (commented):
Hey @dayofdoom ! Thanks for the PR.

I'm curious: what part of this made the Kagi MCP server work? Was it adding support for manually specifying env vars in the `env` dictionary?

In general, I think this PR is probably a little large — I don't think we need to sanitize user-provided MCP server names when running code on the same user's machine, for example. A lot of the sanitization code seems to be redundantly sanitizing things from the config, whereas I think if the config itself is incorrect that's not something Octo should try to auto-fix.

I do want to fix whatever was causing the Kagi MCP server to not work — or at least document how to use it, if there's an existing way to get it to work without changes — so I'm curious which part of this PR got it working for you.

PR #46, opened 2025-10-30
Add functionality to compact messages 

sdjidjev:
Splitting up the auto-compact feature into two PRs for simplicity.
- This PR just implements the functionality of compacting
- The next PR will integrate compacting into octo with auto-compact (and its associated config)

Major changes in this PR:
- Rename `onAutofixJson` callback into `onActivity` to do callbacks for other prompts (like compact)
- Adds prompt for compacting messages
- Helper functions for compacting messages
- Adds `CompactSummaryItem`  (to be integrated next PR)
- Adds `CompactSummaryMessage`
- Mock data for testing
- Tests

Part of [SYN-467](https://linear.app/synthetic-lab/issue/SYN-467/auto-compact-conversation-history-in-octofriend)

### DIFF
--- a/source/prompts/compact-prompt.ts
+++ b/source/prompts/compact-prompt.ts
@@ -1,76 +1,62 @@
 export function compactPrompt(conversationHistory: string) {
   return (
-`
-${conversationHistory}
+`Here is the conversation history to summarize: ${conversationHistory}
 
-That was a conversation history above. Now listen to these instructions:
-You are creating a compressed summary of a conversation history that will replace the original messages in the context window.
+You are creating an internal context summary that will replace previous messages in the conversation history.
 
-**Your role**: You are NOT responding to the user. You are NOT continuing the conversation. You are ONLY creating an internal summary document that will be inserted into the context to preserve information about what has already happened.
+**IMPORTANT**: You are NOT continuing the conversation or responding to the user. You are creating a factual record of what has already happened. The user will likely never see this summary - it exists only to preserve context when the conversation resumes.
 
-**Critical constraints**:
-- This summary will REPLACE all previous messages when the conversation resumes
-- The user will NOT see this summary - it's purely for context preservation
-- You must write in third-person past tense (e.g., "The user requested X", "The assistant implemented Y")
-- Do NOT include greetings, signoffs, or conversational language
-- Do NOT say things like "I'll help you with..." or "Let me know if..."
-- Do NOT attempt to continue or complete any in-progress work
-- Simply document what has already occurred
+**Your task**: Analyze the conversation history that you read at the beginning of this message and create a detailed technical summary capturing all essential information for continuing work.
 
-**Required sections**:
+Before writing your summary, use <analysis> tags to organize your thinking:
 
-## Current State
+<analysis>
+[Chronologically review each part of the conversation and identify:
+- The user's explicit requests and goals
+- Actions that were taken and their outcomes  
+- Technical decisions and their rationale
+- Important code changes, file modifications, and command outputs
+- Problems encountered and their resolutions
+- Work that remains incomplete]
+</analysis>
 
-- What task the user requested (exact wording if possible)
-- What has been completed so far
-- What was in progress when the conversation paused (incomplete/partial work)
-- What was explicitly discussed as remaining work (not your inference, but what was actually mentioned in the conversation)
+Then provide your summary in this structure:
 
-## Files & Changes
+<summary>
+## Primary Request
+[What the user asked for, in their own words if possible]
 
-- Files that were created or modified (with brief description of changes)
-- Files that were read or analyzed (why they were relevant)
-- Key files mentioned but not yet modified
-- Specific file paths and line numbers for important code locations
+## Work Completed
+[Specific actions taken, files modified, commands run - with results]
 
 ## Technical Context
+- Architecture decisions made and why
+- Patterns, frameworks, libraries being used (with examples)
+- Commands that worked (exact syntax + relevant output)
+- Commands that failed (what was tried, error messages, why it failed)
+- Environment details (versions, dependencies, system info)
 
-- Architecture decisions that were made and their rationale
-- Patterns and conventions being followed (with examples from the conversation)
-- Libraries, frameworks, and tools being used
-- Commands that were executed successfully (exact commands with output context)
-- Commands that failed (what was attempted and why it failed)
-- Environment details (language versions, dependencies, system info)
+## Files Modified
+For each file:
+- **[Filename]**: [Why it's relevant] [What changed] [Current state]
 
-## Strategy & Approach
+## Work In Progress
+[Specific description of incomplete work - which files, what state they're in, what was being attempted]
 
-- The overall approach being taken to solve the problem
-- Why this approach was chosen (if discussed)
-- Key insights, discoveries, or gotchas that were identified
-- Assumptions that were made
-- Any blockers, risks, or open questions that were raised
+## Outstanding Issues
+- Errors not yet resolved (include error messages)
+- User requests acknowledged but not completed  
+- Known blockers or risks
 
-## Unfinished Work
+## Context for Resuming
+[The most important things to know to pick up where this left off - reference specific files, line numbers, or error states]
+</summary>
 
-- What was explicitly stated as still needing to be done
-- Work that was started but not completed (with specifics: which file, which function, what state it's in)
-- Issues or errors that were encountered but not yet resolved
-- User requests that were acknowledged but not yet addressed
-
-**Format requirements**:
-- Write in past tense, third person (chronicle what happened)
-- Be factual and detailed, not conversational
-- Think of this as technical documentation, not a message
-- Include specific details: file names, function names, error messages, exact commands
-- Capture the state of the work as it was left, not predictions about what should happen
-
-**Length**: No limit. Prioritize completeness over brevity. Missing context cannot be recovered.
-
-**Format**: You may NOT output in JSON format. You must output in the markdown format outlined above with the specified required sections.
-
-Here is that same conversation history that you are summarizing, read the following very carefully:
-
-${conversationHistory}
+**Formatting rules**:
+- Write in past tense ("The user requested...", "The assistant modified...", "The build failed...")
+- Be specific: include file paths, function names, exact error messages, command syntax
+- Capture enough detail that work could resume without re-reading the full history
+- This is technical documentation, not a conversation
 `
   );
 }


source/compilers/anthropic.ts
```diff
@@ -149,6 +150,13 @@ async function modelMessageFromIr(
     };
   }
 
+  if(ir.role === "compact-summary") {
+    return {
+      role: "assistant",
```
sdjidjev: i'm not 100% sure if an `assistant` role should be chosen for a `compact-summary`, but it seems to be the most appropriate 🤔 would love a double-check on this!
(same comment goes for all the other compilers)

source/app.tsx
```diff
@@ -615,8 +616,9 @@ const MessageDisplayInner = React.memo(({ item }: {
     return <Text color="red">Request failed.</Text>
   }
 
+  // TODO: Add rendering for compact-summary when autoCompact feature is integrated
   // Type assertion proving we've handled all types other than user
```
billycao: super-nit: could prob remove this comment. seems clear enough and we don't use it on line 279 anyways :P (edit: the comment that was already there, not your comment :P)

source/compilers/autocompact.ts
```diff
@@ -0,0 +1,98 @@
...
+import { toLlmIR } from "../ir/llm-ir.ts";
+import { countIRTokens } from "../ir/ir-windowing.ts";
+import { compactPrompt, CompactResponse } from "../prompts/compact-prompt.ts";
+import { ActivityMode } from "../state.ts";
+import { Transport } from "../transports/transport-common.ts";
+import { run } from "./run.ts";
+
+export function shouldCompactHistory(
+  history: HistoryItem[],
+  config: Config,
```
billycao, line 66: Instead of being defined globally, we should probably use `modelConfig.context` like the existing windowing functionality uses in line 30 of `source/compilers/run.ts`.

In fact that file would probably be a sensible entrypoint for auto-compaction

source/compilers/autocompact.ts
```diff
@@ -0,0 +1,98 @@
...
+): boolean {
+  if (!config.autoCompact?.enabled) return false;
```
billycao: I think this should default to true so users benefit from auto-compaction without enabling it, and can explicitly disable it if they wanted.

source/prompts/compact-prompt.ts
```diff
@@ -0,0 +1,33 @@
...
+Respond with JSON in the following format:
+
+// Success response:
+${toTypescript({ CompactSuccess })}
+
+// Failure response (if the history cannot be meaningfully summarized):
+${toTypescript({ CompactFailure })}
+
+Here's the conversation history to summarize:
+${conversationHistory}`
```
billycao: I'm not an expert on tokenizers and how they handle system/user/assistant prompts (and whether there's any performance uplift), but I think this could be formatted as a system prompt "You are an expert summarizer of coding agent conversation histories. The user will provide you with the conversation history between a user and a coding agent in JSON format. Your task [...]", and then the user prompt can contain the history.

That can just be a string system prompt which is directly exported, kind of like how `system-prompt.ts` is used.

source/compilers/anthropic.ts
billycao: My 2c is that instead of replacing the history with an assistant message, the history (at least the fact that the history was compacted) can be provided as part of the system prompt, the same way `appliedWindow` is currently passed to `systemPrompt()` in `source/prompts/system-prompt.ts`.

ie. The entrypoint can be `source/compilers/run.ts`, where we conditionally `applyContextWindow` or `applyCompaction`, and that can be passed to systemPrompt in each compiler. 

`systemPrompt()` can conditionally include a system message like "Due to context length limits, part of the conversation history has been summarized. Treat the conversation history as if it was part of the current conversation thread.

This is the conversation history summary: ${summary}"

source/compilers/autocompact.ts
sdjidjev: sg! i'm going to put a todo here for now.
i just want it to stay `false` until i release the config PR is released so it doesn't accidentally get triggered before then.

source/prompts/compact-prompt.ts
sdjidjev: Oh I think I may be misunderstanding your comment-- I was trying to model it after `system-prompt.ts` and `autofix-prompts.ts` when i wrote this; by taking in the conversation history as a string and then telling it to format the response with JSON.

Not sure what to do differently here, would love a little more guidance!

billycao: Whoops, it probably would have been more clear if I put my comment in `source/compilers/autocompact.ts` :P

My suggestion is to rename `compactPrompt()` to `compactSystemPrompt()` and provide it in a system message, and provide the history as user message. That being said I'm not 100% on inference engine implementations, but I do have some intuition that system messages are better adhered to than user messages.

Also part of the motivation is separation of concerns for easier changes in the future.

(though I do realize splitting out the system prompt requires updating the call stack for `runInternal`, so feel free to ignore. :)

==

One small note, I notice in `source/compilers/run.ts:54` we `applyContextWindow()` to the output of `formatMessagesForSummary()`, but doesn't that fn always only ever return 1 very long user message? If I'm understanding correctly that will either never do anything or always return error since it only operates at the IR granularity.

I think one possible implementation would be to

```typescript
  if (shouldAutoCompactHistory(messages, modelConfig.context, config.autoCompact)) {
    const systemPromptIR = compactSystemPromptIR();  # to implement
    const windowedIR = applyContextWindow([ systemPromptIR, ...messages], modelConfig.context);  # update applyContextWindow to always include the first message if it is a system message and fits in context.
    
    const windowedIR = [ systemPromptIR, formatMessagesForSummary(windowedIR.ir.slice(1)) ]  # update formatMessagesForSummary to omit the system prompt and just return the summary as a user message
```

This way we properly window the history to the context limit. Hope that makes sense! :) Sorry if this breaks your existing unit tests! 😅 

That being said one downside of this implementation would be that it it requires modifying the `LlmIR` and `applyContextWindow` interface to include system messages, and I'm not sure if Matt intended for those to always be non-system messages.

One alternative is to ignore the system message as part of `applyContextWindow`, and instead just `countIRTokens(systemPromptIR)` and call `formatMessagesForSummary(applyContextWindow(messages, modelConfig.context - countIRTokens(systemPromptIR)).ir)` to get the summary, though I'm a proponent of improving `applyContextWindow`

Hope what I said made sense! Tbh I'm not an expert on the octofriend codebase so if I'm misunderstand please feel free to ignore me. But I do think the context window contraction bug should be fixed, if I'm reading this code right :)

sdjidjev: Yeah great catch with the applyContextWindow stuff, definitely gonna refactor that with your idea in mind!

source/compilers/run.ts
```diff
@@ -19,16 +40,58 @@ export async function run({
   skipSystemPrompt?: boolean,
 }) {
   const modelConfig = getModelFromConfig(config, modelOverride);
-  const run = (() => {
+  const runInternal = (() => {
```
reissbaker: Architecture feedback — my high-level idea of what `run` should do is:

1. Take `LLMIR`
2. Run it more-or-less unmodified

Now, it doesn't actually work this way entirely; in fact what it does is kind of annoying:

1. Take `LLMIR`
2. Generate a system prompt unless you tell it not to
3. Run the system prompt + LLMIR
4. Call autofix stuff if the response is broken

I think (4) is actually reasonable, but (3) is obviously not just a straightforward transformation. Probably we should remove it (not in this PR! just in general): there's already the code smell of `skipSystemPrompt?: boolean`, which sort of implies that we should be passing in the system prompt itself as LLMIR rather than having the compilers secretly insert it on our behalf.

I think the autocompaction should probably work in the more Platonic ideal sense, where it's something that happens outside of the `run` function, for a couple reasons:

1. IDK, it feels cleaner
2. If we add another feature that needs to call an LLM, it will have to use `run`, and it's a lot to hold in your head that `run()` might try to autocompact (and also might insert a system prompt)

So I think it'd be nice to make the "run with autocompaction" function a separate function which calls `run`, and keep `run` itself relatively minimal. Perhaps the new function can be caled `agentRun`? And then in `state.ts`, we'd update the calls to `run` to instead call `agentRun`.

I think we could also make `agentRun` no longer have an option to `skipSystemPrompt`; `agentRun` would basically be "Run this as Octo, i.e. with its system prompt, autocompaction, etc", vs `run` which simply wraps LLM APIs and converts LLMIR to the underlying correct API format.

source/compilers/run.ts
```diff
@@ -19,16 +40,58 @@ export async function run({
...
     const _: "anthropic" = modelConfig.type;
     return runAnthropicAgent;
   })();
 
-  const windowedIR = applyContextWindow(messages, modelConfig.context);
+  let compactSummary: string | undefined;
+  let windowedIR: WindowedIR;
+  let processedMessages = messages
+
+  if (shouldAutoCompactHistory(messages, modelConfig.context, config.autoCompact)) {
```
reissbaker: So, I think there's a subtle issue here:

1. This runs autocompaction whenever the context is longer than some threshold (and autocompaction is turned on in the config, but let's ignore that for now)
2. But, we never return the results of the autocompaction to the `state.ts` caller once we've run autocompaction — so, once we pass the critical threshold, the context size of our history array never shrinks (we keep appending responses to it), and we'll run autocompaction on every turn. This will burn API requests, and also probably make it harder for the LLM to keep track of what it's doing; compaction is a pretty brutal process that strips a lot of context out.

For example:

```
1. we have history [a, b, c, d]
2. we autocompact to [abcd-summary]
3. we return the response, e
4. next turn: we have the history [a, b, c, d, e]
5. we autocompact to [abcde-summary]
6. etc
```

I think we actually need a new kind of LLMIR! Probably: `CompactionCheckpoint`:

```typescript
type CompactionCheckpoint = {
  role: "compaction-checkpoint",
  summary: string,
};
```

In `state.ts`, at the top of `_runAgent`, we'd do the following:

1. Check `shouldAutocompact`
2. If we should autocompact, generate a `CompactionCheckpoint` and add it to our history array
3. Then, continue with the rest of the function

Then in our `run` function, before passing the LLMIR off we would:

1. Loop through the LLMIR backwards, to find the index of the most-recent `CompactionCheckpoint` (stopping at index 0)
2. `messages.slice(compactionCheckpointIndex)` to make `run` only consider the most-recent messages since the checkpoint.

This means that once we run autocompaction once, in theory we shouldn't have to run it again for a long time — since we store the checkpoints in history, we don't need to keep re-checkpointing. AKA:

```
1. we have history [a, b, c, d]
2. we autocompact to abcd-summary and add it to history: [a, b, c, d, abcd-summary]
3. we call run()
4. run() slices to [abcd-summary] 
5. run() returns the response, e
6. next turn: we have the history [a, b, c, d, abcd-summary, e]
7. we call run()
8. run() slices to [abcd-summary, e]
9. we do not need to autocompact again yet
```

Then we'd make `systemPrompt(...)` take an optional compaction checkpoint. If the IR starts with a compaction checkpoint, we'd add it to the system prompt, and we wouldn't need the special flags of `appliedCompaction` or `compactSummary`: if the LLMIR starts with a `CompactionCheckpoint` it clearly applied compaction (and the summary is contained in the checkpoint); otherwise, it didn't.

At a high level I think this also keeps `run` as a cleaner API that mostly just wraps LLM APIs, rather than something that potentially rather deeply modifies the passed-in LLM IR behind the scenes by autocompacting. This is also nice just as a sanity check for using `run(...)` internally, since you know it more-or-less just executes the LLMIR you give it (albeit the system prompt insertion stuff is annoying and I probably shouldn't have put that into `run` in the first place).

And in this world, since `CompactionCheckpoints` are `LLMIR` in the history array, we can also then render the compaction checkpoint in the UI! So that users know and see that compaction happened.

There's one bit of nuance here: I think some LLM APIs will not like completions that start from scratch where there isn't a user message to respond to. So you'll probably need to add some sort of generated user message to the history array, telling the LLM to check the system prompt for its compaction information and continue with its task. AKA:

```
1. we have history [a, b, c, d]
2. we autocompact to abcd-summary and add it to history: [a, b, c, d, abcd-summary]
3. we add a generated user message: [a, b, c, d, abcd-summary, fake-user-message]
4. we call run()
5. run() slices to [abcd-summary, fake-user-message] 
6. run() returns the response, e
7. next turn: we have the history [a, b, c, d, abcd-summary, fake-user-message, e]
8. we call run()
9. run() slices to [abcd-summary, fake-user-message, e]
10. we do not need to autocompact again yet
```

source/compilers/autocompact.ts
```diff
@@ -0,0 +1,64 @@
...
+export function formatMessagesForSummary(messages: LlmIR[]): LlmIR[] {
+  const lines: string[] = [];
+
+  for (const message of messages) {
+    // Create a copy without the id field and stringify
+    const { role, ...rest } = message;
+    lines.push(`${message.role}: ${JSON.stringify(rest)}`);
+  }
+
+  const conjoinedMessages = lines.join("\n");
```
reissbaker, line 107: One note that will help LLMs handle compaction more correctly: agents might be in the middle of handling a tool call failure when this gets called; if we keep the failures in, it will probably show up in the summary results, which will somewhat mislead LLMs (you probably wouldn't instruct an LLM with "and here are some errors"). It's probably worth filtering out any malformed tool calls or tool call failures from getting summarized.

source/prompts/compact-prompt.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function compactPrompt(conversationHistory: string) {
+  return (
+`You are helping to compact a conversation history to save context tokens. Your task is to create a concise summary of the conversation history provided below that preserves all critical information while being as brief as possible.
+
+The summary should:
+- Preserve key decisions, implementations, and outcomes
+- Include important context that might be needed for future messages
+- Maintain technical details that could be referenced later
+- Be written in a clear, factual style
```
reissbaker: I think there's also one more thing it should summarize: what the current goal is! For reference, this is Crush's summary prompt: https://github.com/charmbracelet/crush/blob/303bf8c846660c4a7c295a0a154354d8f0c4413b/internal/agent/templates/summary.md

FWIW, I'm not sure if it's worth forcing the LLM to respond in JSON and allowing failure. LLMs are very good at summarizing, and summarizing in plain text is probably more in-distribution (aka, it was trained on text completions like this) rather than responding with a summary wrapped inside JSON. A weird but helpful-to-know fact: early versions of LLMs were actually better at writing code not-inside-tool-calls, because they saw lots of code in their training runs but didn't see lots of code wrapped inside JSON in their training runs. All of the major labs have now fixed this, probably via synthetic data generation where they wrap code edits inside JSON during training runs (so models are now good at generating code inside JSON, since they see it in training), but I suspect they haven't done this for summaries since they're pretty focused on tool calls for coding agents rather than summarization.

You can probably just ask the LLM to summarize the text, and take its output as the summary.

source/compilers/autocompact.ts
sdjidjev: @reissbaker i was thinking about filtering it out earlier but there are sometimes tool call failures are actually kind of a good/expected thing? like in this example:
Screenshot 2025-11-24 at 8 48 31 PM
should we then keep tool call failures at least? i'm not sure of instances where malformed tool calls would be an expected thing too but i haven't come across any

source/compilers/run.ts
sdjidjev: @reissbaker thank you for the detailed guidance!
question: i'm a little confused in your description if `abcd-summary` is just the prompt to make `e` (summarizing a thru d) and `fake-user-message` is just a static string telling it to check the system prompt? or is `e` the output of user-prompt `d` + `abcd-summary` (where `abcd-summary` is the result of the compaction)?

source/compilers/autocompact.ts
reissbaker: Oh, good point — let's just avoid filtering! Since the LLM might respond to a malformed tool call by thinking about how it was malformed, it's probably just less confusing to keep everything in.

source/compilers/run.ts
reissbaker: Oh, `abcd-summary` is the *output* of the compaction! So i.e.


1. [user-msg, assistant-msg, tool-output, assistant-msg, tool-output]`
2. we autocompact to a summary `summary-result-checkpoint` by calling your autocompaction code, saving the summary into the checkpoint. now we have [user-msg, assistant-msg, tool-output, assistant-msg, tool-output, summary-result-checkpoint`]
3. we add a fake user message like "keep working on your task" and have [user-msg, assistant-msg, tool-output, summary-result-checkpoint, fake-user-message]
4. we call `run()`
5. `run()` slices to [summary-result, fake-user-message]
6. `run()` returns a new assistant-msg
7. next turn: we have the history [user-msg, assistant-msg, tool-output, assistant-msg, tool-output, summary-result-checkpoint, fake-user-message, assistant-msg]
8. we call `run()`
9. `run()` slices to [summary-result-checkpoint, fake-user-message, assistant-msg]
10. we do not need to autocompact yet

Basically, our `state.ts` never alters the history, it just keeps appending to it. So, we return checkpoints that include the autocompaction summary, and in `run()` we slice so that only the most-recent checkpoint (and any subsequent messages) are included in the LLMIR. In the various compilers, we insert the checkpoint summary into the system prompt (if a checkpoint summary exists).

sdjidjev:
@reissbaker

## Open questions:
1. Should we allow configuring the autocompact threshold? I set it to a very low amount (0.01) for testing and octo behaved very..... strangely lol. Claude has it preset to something that you can't change.... kinda seems like the move 🤔 
2. Right now I inserted a fake user message after compaction that says "Continue working on your task" like we talked about, but it does look sort of strange in the UI, as it looks like the user wrote it even though they didn't... Should I create a new message type like `IntermediatePromptMessage` or something and keep it hidden? Or should I add a line to the end of the compaction-checkpoint summary telling it to continue working on its task?

## TODO:
I wanted to write a more comprehensive test to make sure the compaction-checkpoint & context window slicing works as expected but other than that everything seems to be working. Was gonna write that tomorrow but in the meantime would be great to have you review!!

### Config screenshots
|  |  |
| --- | --- |
|  |  |

### In-progress compaction:
Screenshot 2025-12-18 at 5 31 04 PM

### Post-compaction
Screenshot 2025-12-18 at 5 30 31 PM

PS I think octo is onto me.....
Screenshot 2025-12-18 at 12 31 00 AM

reissbaker:
Will do a full review in the am, but very exciting! Re: open questions —

1. I think a static percentage as a default seems right. We can always add configuration in the future if we feel like we need it.
2. Oh yeah, definitely showing the fake message in the UI seems confusing. Thinking about this more: what if we handle this in the compilers specifically? I'm actually not sure how well every LLM will handle conversations that *start* with an assistant message, so potentially we could just silently compile the entire compaction checkpoint to a user message like `We had a previous conversation that was getting long, so you generated a summary:<octo-summary>insert summary here</octo-summary>Now, keep doing what you were doing previously` or sth like that? Then the fake user message doesn't live in the UI, it just lives in the compiled messages.

source/menu.tsx
```diff
@@ -161,6 +164,171 @@ function FixJsonToggle() {
...
+  const setConfig = useSetConfig();
+  const { setMenuMode } = useMenuState(useShallow(state => ({
+    setMenuMode: state.setMenuMode,
+  })));
+  const { toggleMenu, notify } = useAppStore(useShallow(state => ({
+    toggleMenu: state.toggleMenu,
+    notify: state.notify,
+  })));
+
+  const [mode, setMode] = React.useState<"menu" | "edit-threshold">("menu");
```
reissbaker: style nit — probably import `useState` at the top and just use `useState<...>(...)` rather than `React.useState<...>(...)`

source/state.ts
```diff
@@ -225,68 +232,146 @@ export const useAppStore = create<UiState>((set, get) => ({
...
+
+        compactionUpdater.clear();
+
+        if (checkpointSummary) {
+          const checkpointItem: CompactionCheckpointItem = {
+            type: "compaction-checkpoint",
+            id: sequenceId(),
+            summary: checkpointSummary,
+          };
+          const continuationItem: UserItem = {
```
reissbaker: Like we chatted about — I think actually we can skip this, and just have the compilers include the line `Continue working on your task` or equivalent

source/state.ts
```diff
@@ -225,68 +232,146 @@ export const useAppStore = create<UiState>((set, get) => ({
...
+            type: "user",
+            id: sequenceId(),
+            content: "Continue working on your task.",
+          };
+          historyCopy.push(checkpointItem);
+          historyCopy.push(continuationItem);
+          set({ history: historyCopy });
+        }
+      } catch (e) {
+        get().notify("Autocompaction failed: " + e);
```
reissbaker: IMO this is a hard failure and we shouldn't catch it and move on unless it's an explicit abort (since a compaction request failure means the model will run out of context space, which will cause future failures) — probably worth setting some state to show an error message, bailing, and in the UI show the error + option to retry, like we do with regular assistant message failures.

Also, probably worth bailing even if it's an abort, to return control to the user — just don't need to show an error message in that case.

source/state.ts
```diff
@@ -225,68 +232,146 @@ export const useAppStore = create<UiState>((set, get) => ({
...
+    const onAutofixJson = () => { set({ modeData: { mode: "fix-json" } }); };
+
+    const debounceTimeout = 100;
+    const compactionAbortController = new AbortController();
+    const agentAbortController = new AbortController();
+
+    const createDebouncedUpdater = () => {
+      let timeout: NodeJS.Timeout | null = null;
+      return {
+        schedule: (updateFn: () => void) => {
```
reissbaker: oo very nice

source/compilers/autocompact.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+export function shouldAutoCompactHistory(
+  messages: LlmIR[],
+  onError: (errString: string) => void,
+  config: Config,
+  modelOverride: string | null,
+  autoCompactSettings?: AutoCompactConfig,
+): boolean {
+  if (!autoCompactSettings?.enabled) return false;
```
reissbaker: IMO autocompact should default to on — so we'd probably want:

```typescript
if(autoCompactSettings != null && !autoCompactSettings.enabled) return false;

const contextThreshold = autoCompactSettings?.contextThreshold || 0.9; // Claude defaults to 0.9 so idk seems good
```

source/prompts/compact-prompt.ts
```diff
@@ -0,0 +1,64 @@
...
+  return (
+`You are creating an internal context summary that will replace previous messages in the conversation history.
+
+**IMPORTANT**: You are NOT continuing the conversation or responding to the user. You are creating a factual record of what has already happened. The user will likely never see this summary - it exists only to preserve context when the conversation resumes.
+
+**Your task**: Analyze the conversation history and create a detailed technical summary capturing all essential information for continuing work.
+
+Before writing your summary, use <analysis> tags to organize your thinking:
+
+<analysis>
```
sdjidjev, line 13: This prompt is a hybrid of [Crush's](https://github.com/charmbracelet/crush/blob/main/internal/agent/templates/summary.md) and [Claude's](https://github.com/Piebald-AI/claude-code-system-prompts/blob/main/system-prompts/agent-prompt-conversation-summarization.md) compaction prompts, plus some reinforcement that the output is meant to be for summarization, not for a user to read.

PR #53, opened 2025-11-20
Implement basic Vim mode support

bradfa:
Add foundational Vim editing capabilities to allow users to toggle between Emacs (default) and Vim modes. This implementation provides modal editing with NORMAL and INSERT modes.

Key features:
- Vim state management integrated into Zustand store with vimEnabled and vimMode fields
- Config schema extended to persist vimEnabled preference
- Mode indicator displays "--INSERT--" in green text
- Basic Vim commands: i (enter INSERT), a (enter INSERT after cursor), h/l (move left/right), b/w/e (move by word), 0/$/^ (move across line), A/I (enter INSERT at end/start of line), x (delete char)
- Context-aware Esc handling: INSERT mode Esc→NORMAL, other cases→menu
- Full backward compatibility with Emacs mode unchanged
- Exiting octo still ctrl+c ctrl+c or from menu

### DIFF
--- a/source/app.tsx
+++ b/source/app.tsx
@@ -43,6 +43,7 @@ import { useCtrlC, ExitOnDoubleCtrlC, useCtrlCPressed } from "./components/exit-
 import { InputHistory } from "./input-history/index.ts";
 import { Markdown } from "./markdown/index.tsx";
 import { countLines } from "./str.ts";
+import { VimModeIndicator } from "./components/vim-mode.tsx";
 
 type Props = {
 	config: Config;
@@ -195,6 +196,8 @@ function BottomBarContent({ inputHistory }: { inputHistory: InputHistory }) {
   const config = useConfig();
   const transport = useContext(TransportContext);
 	const [ query, setQuery ] = useState("");
+  const [vimEnabled, setVimEnabled] = useState(config.vimEmulation?.['enabled'] ?? false);
+  const [vimMode, setVimMode] = useState<'NORMAL' | 'INSERT'>('NORMAL');
   const { modeData, input, abortResponse, toggleMenu, byteCount } = useAppStore(
     useShallow(state => ({
       modeData: state.modeData,
@@ -205,13 +208,27 @@ function BottomBarContent({ inputHistory }: { inputHistory: InputHistory }) {
     }))
   );
 
+  // Sync local vimEnabled state when vimEmulation config changes
+  useEffect(() => {
+    if (config.vimEmulation?.['enabled'] !== undefined) {
+      setVimEnabled(config.vimEmulation['enabled']);
+    }
+  }, [config.vimEmulation]);
+
   useCtrlC(() => {
     setQuery("");
   });
 
   useInput((_, key) => {
     if(key.escape) {
-      abortResponse();
+      // Vim INSERT mode: Esc ONLY returns to NORMAL (no menu, no abort)
+      if (vimEnabled && vimMode === 'INSERT') {
+        setVimMode('NORMAL');
+        return;
+      }
+
+      // All other cases: abort response (if active) and open menu
+      abortResponse();  // Safe to call even if no response is active
       toggleMenu();
     }
   });
@@ -286,7 +303,11 @@ function BottomBarContent({ inputHistory }: { inputHistory: InputHistory }) {
       value={query}
       onChange={setQuery}
       onSubmit={onSubmit}
+      vimEnabled={vimEnabled}
+      vimMode={vimMode}
+      setVimMode={setVimMode}
     />
+    <VimModeIndicator vimEnabled={vimEnabled} vimMode={vimMode} />
   </Box>
 }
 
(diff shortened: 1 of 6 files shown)


source/app.tsx
```diff
@@ -89,6 +90,13 @@ export default function App({ config, configPath, metadata, unchained, transport
     }))
   );
 
+  // Initialize vim state from config on first render
+  const vimInitialized = useRef(false);
+  if (!vimInitialized.current && config.vimEnabled !== undefined) {
+    useAppStore.getState().setVimEnabled(config.vimEnabled);
```
reissbaker: Rather than ferrying data at startup from the config to Zustand, I think if a component needs to know whether Vim mode is enabled, it should check the config itself via:

```typescript
const config = useConfig();
// Callers can access config.vimEnabled directly
```

The nice thing about doing it this way is that `useConfig` already tracks any config changes made by the current Octo process — so for example, in your menu code, you won't have to set the config *and* call `toggleVimMode()`: you can instead just set the config.

source/app.tsx
```diff
@@ -211,7 +222,14 @@ function BottomBarContent({ inputHistory }: { inputHistory: InputHistory }) {
 
   useInput((_, key) => {
     if(key.escape) {
-      abortResponse();
+      // Vim INSERT mode: Esc ONLY returns to NORMAL (no menu, no abort)
+      if (vimEnabled && vimMode === 'INSERT') {
```
reissbaker, line 225: Ditto here — this component actually already loads the config via `useConfig`, so it could instead do:

```typescript
if(config.vimEnabled && vimMode === "INSERT") {
  // ...
}
```

source/config.ts
```diff
@@ -43,6 +43,7 @@ const ConfigSchema = t.exact({
     apiEnvVar: t.optional(t.str),
     model: t.str,
   })),
+  vimEnabled: t.optional(t.bool),
```
reissbaker: A bit of a bikeshed, but — I think it would be nice to call this `vimEmulation`, and have it be:

```typescript
vimEmulation: t.optional(t.subtype({
  enabled: t.bool,
})),
```

Mainly because I think it'll be nice to extend this in the future to support customizing some of the Vim keybindings. For example, I have some custom keybindings I often use for Vim where `L` (capitalized) is equivalent to `$`, and `H` (capitalized) is equivalent to `^`. Since these are non-standard it wouldn't make sense to put them into Octo directly, but it would be nice to allow users (such as myself!) to customize them in the future.

Not worth implementing custom keybindings in this PR directly (I think it would make the PR too big), but it'll be nice to leave the option to add it in the future by making this config format change!

source/menu.tsx
```diff
@@ -313,7 +321,12 @@ function MainMenu() {
       value: "quit" as const,
     },
   ];
+
   items = items.filter(item => {
+    // vim toggle should always be visible
+    if(item.value === "vim-toggle") return true;
```
reissbaker: You can delete this line — line 332 already always returns `true` if the other menu items aren't applicable :)

source/menu.tsx
```diff
@@ -324,11 +337,27 @@ function MainMenu() {
...
-    if(item.value === "return") toggleMenu();
-    else if(item.value === "quit") setMenuMode("quit-confirm");
+  const onSelect = useCallback(async (item: (typeof items)[number]) => {
+    if (item.value === "return") toggleMenu();
+    else if (item.value === "quit") setMenuMode("quit-confirm");
+    else if (item.value === "vim-toggle") {
+      const wasEnabled = vimEnabled;
+
+      // Update Zustand state
+      toggleVimEnabled();
```
reissbaker: If you remove the `vimEnabled` part from Zustand, you can delete this line since everyone will just read it transparently from the config :)

You may still want to set Vim mode to `NORMAL` here, though, since I notice that `toggleVimEnabled` does both of those.

source/components/vim-mode.tsx
```diff
@@ -0,0 +1,228 @@
...
+  vimMode
+}: {
+  vimEnabled: boolean;
+  vimMode: 'NORMAL' | 'INSERT';
+}) {
+  if (!vimEnabled || vimMode === 'NORMAL') return null;
+
+  return (
+    <Box>
+      <Text color="green" bold>-- INSERT --</Text>
```
reissbaker: Minor nit: I think this would look nicer if you used the Octo theme colors. You can do this by:

```tsx
import { useColor } from "../theme.ts";

// In your component:
const color = useColor();

// Later:
return (
  <Box>
    <Text color={color}>-- INSERT --</Text>
  </Box>
);
```

reissbaker (changes requested):
As a major Vim addict, I am a huge fan of this change!

I left a few comments on the implementation to address. Nothing major, overall this looks quite nice, just some improvements on integrating with Octo's codebase.

source/components/vim-mode.tsx
bradfa: Yes, sorry, didn't realize how to use the theming.  Will fix.

source/app.tsx
bradfa: Gotcha!  I didn't fully understand the difference between the Zustand store and the config.  Will fix.

PR #58, opened 2025-12-03
Combat flickering in streaming text by using a ScrollView

sdjidjev:
### Before
Flickering behavior before ScrollView implementation in iTerm:
<video src='https://github.com/user-attachments/assets/92039098-8975-45ef-a8e3-5c6e1633603a' width=180/>

### After
Streaming content inside a ScrollView doesn't flicker in iTerm! 🎉 
<video src='https://github.com/user-attachments/assets/910e220a-1535-4215-bee3-084a15361943' width=180/>

@reissbaker 
@billycao

### DIFF
--- a/source/components/scroll-view.tsx
+++ b/source/components/scroll-view.tsx
@@ -42,15 +42,19 @@ export interface ScrollViewProps extends React.PropsWithChildren {
 export default function ScrollView({ height, children }: ScrollViewProps) {
   const [innerHeight, setInnerHeight] = useState(0);
   const [scrollTop, setScrollTop] = useState(0);
+  const [shouldAutoScroll, setShouldAutoScroll] = useState(true);
   const innerRef = useRef<DOMElement>(null);
   const { stdin, setRawMode } = useStdin();
-  const { stdout } = useStdout();
 
-  const measureInnerElement = useCallback(() => {
+  const handleElementSize = useCallback(() => {
     if (!innerRef.current) return;
     const dimensions = measureElement(innerRef.current);
     setInnerHeight(dimensions.height);
-  }, []);
+    if (shouldAutoScroll) {
+      const maxScroll = Math.max(0, dimensions.height - height);
+      setScrollTop(maxScroll);
+    }
+  }, [shouldAutoScroll]);
 
   const isScrollable = innerHeight > height;
   const maxScroll = Math.max(0, innerHeight - height);
@@ -61,20 +65,23 @@ export default function ScrollView({ height, children }: ScrollViewProps) {
       const delta = direction === SCROLL_DIRECTIONS.SCROLL_UP ? -1 : 1;
       const newScrollTop = prev + delta;
       const maxScroll = Math.max(0, innerHeight - height);
+      const scrollPercentage = maxScroll > 0 ? Math.round((newScrollTop / maxScroll) * 100) : 0;
+      if (scrollPercentage >= 99) setShouldAutoScroll(true);
+      else setShouldAutoScroll(false);
       return Math.max(0, Math.min(newScrollTop, maxScroll));
     });
   }, [innerHeight, height]);
 
   useEffect(() => {
     const handleResize = () => {
-      setTimeout(measureInnerElement, 0);
+      setTimeout(handleElementSize, 0);
     };
     process.stdout.on('resize', handleResize);
 
     return () => {
       process.stdout.off('resize', handleResize);
     };
-  }, [measureInnerElement]);
+  }, [handleElementSize]);
 
   useEffect(() => {
     if (!stdin || !setRawMode || !isScrollable) {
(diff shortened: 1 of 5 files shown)


source/components/scroll-view.tsx
```diff
@@ -0,0 +1,181 @@
+import { Box, DOMElement, measureElement, useStdin, Text, BoxProps, useStdout } from 'ink';
+import React, { useEffect, useState, useRef, useCallback } from 'react';
+
+// https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Extended-coordinates 
+const MOUSE_PATTERNS = {
```
sdjidjev, line 5: right now I'm manually grabbing the mouse events by regex and hardcoded escape sequences. i don't think `ink` exposes mouse events.
idk how conservative we want to be with installing libraries but [terminal-kit](https://www.npmjs.com/package/terminal-kit) seems way more robust and it can handle mouse scroll events! 🐭 
i feel like it has other features that we could leverage later on!

reissbaker: Oh wow, terminal-kit looks nice! For now we can keep using your existing regexes, but in the long run it would be nice to add it and handle e.g. clickable buttons via a mouse.

source/components/scroll-view.tsx
```diff
@@ -0,0 +1,181 @@
...
+    borderColor: SCROLL_UI_COLOR,
+  };
+
+  const borderOffset = (scrollPercentage == 0 ? 1 : 0) + (scrollPercentage == 100 ? 1 : 0);
+  const effectiveHeight = height + borderOffset;
+
+  return (
+    <Box flexDirection="column">
+      <Box
+        height={effectiveHeight}
```
reissbaker: Rather than setting the height to the `effectiveHeight` always, I think you want two changes:

1. Actually, the max height is `effectiveHeight - 1`! You'll still get some flickering with `effectiveHeight` if you fully scroll to the bottom; oddly, you need one row at the top of the screen untouched for re-renders to 100% avoid flicker.
2. You actually probably want to treat this as a max height, rather than always taking up the full screen height — this will help with assistant rendering, and it's also just nice to be able to see what the assistant is doing if e.g. it's doing a bunch of sequential tool calls (otherwise, they'll immediately be pushed offscreen). Although Ink doesn't actually have `maxHeight` support, you can fake it — check out how the `AssistantMessageRenderer` manages the height of the thought box, which expands up until a set maximum:

```typescript
  const [ thoughtsHeight, setThoughtsHeight ] = useState(0);
  const thoughtsOverflow = thoughtsHeight - (MAX_THOUGHTBOX_HEIGHT - 2); // in your case, you'd use the effectiveHeight - 1
  useEffect(() => {
    if(thoughtsRef.current) {
      const { height } = measureElement(thoughtsRef.current);
      setThoughtsHeight(height);
    }
  }, [ thoughts ]);

  // later in the <Box:
            height={thoughtsOverflow > 0 ? MAX_THOUGHTBOX_HEIGHT : undefined}
```

This way you only take up the full height of the terminal if the content starts to overflow.

reissbaker (changes requested):
The scrolling box looks and functions really well! Good job. Terminal scrolling can be pretty annoying to manage. Very impressed with all the different terminal handling you've done + manpage links.

Trying this out on my machine, a couple of notes that didn't fit neatly in line comments:

### Scroll events sometimes leak into the textbox for a second or two

This seems like a frustrating bug, since you actually do call `setRawMode(false)` in the `useEffect` unmount hook. I assume this is React annoyingly reordering hook calls behind the scenes and rendering the text input before calling the `useEffect` unmount. I think there's a simple, slightly hacky workaround: in the `text-input.tsx`, make sure to also set raw mode to false, and ignore input until we know that the raw mode was actually turned off:

```typescript
  const { stdin, setRawMode } = useStdin();
  const [rawOff, setRawOff] = useState(false);

  useEffect(() => {
    setRawMode(false);
    setRawOff(true);
  }, []);

  // Later:
    useInput(
      (input, key) => {
        if(!rawOff) return;
        // rest of existing useInput handler below
```

### Assistant reasoning

I noticed that assistant reasoning seems to get forced offscreen immediately. Using pseudo-max-height as described in line comments would probably fix this though.

### Autoscroll

I think it would be nice to automatically scroll on new assistant input. You'd want to track whether the user is "pinned" to the bottom of the scroll box, so that they can scroll up and have auto-scroll stop, and then scroll down and have autoscroll continue. Hacking around locally, something like this would work:

```typescript
  const [pinned, setPinned] = useState(true);

  // later, add to measureInnerElement (possibly worth renaming to handleElementSize or sth since it now does more than just measuring):
  const measureInnerElement = useCallback(() => {
    if (!innerRef.current) return;
    const dimensions = measureElement(innerRef.current);
    setInnerHeight(dimensions.height);
    // add these lines
    if(pinned) {
      const maxScroll = Math.max(0, dimensions.height - height);
      setScrollTop(maxScroll);
    }
    // make sure to add pinned to the dependency array
  }, [pinned]);

  // later, in handleScroll:
  const handleScroll = useCallback((direction: ScrollDirection) => {
    setScrollTop(prev => {
      const delta = direction === SCROLL_DIRECTIONS.SCROLL_UP ? -1 : 1;
      const newScrollTop = prev + delta;
      const maxScroll = Math.max(0, innerHeight - height);
      // add something like this, where if it's within some delta of the bottom, we re-pin or unpin autoscroll
      const scrollPercentage = maxScroll > 0 ? Math.round((newScrollTop / maxScroll) * 100) : 0;
      if(scrollPercentage >= 99) setPinned(true);
      else setPinned(false);
      return Math.max(0, Math.min(newScrollTop, maxScroll));
    });
  }, [innerHeight, height]);
```

source/components/scroll-view.tsx
sdjidjev: RE: your second point, `to be able to see what the assistant is doing if e.g. it's doing a bunch of sequential tool calls (otherwise, they'll immediately be pushed offscreen)`
I've only encountered where the assistant message & tool calls, etc. are separate. Can tool calls happen in the middle of an assistant message? I Would love to test the behavior out on that if so, lmk what prompts you use to trigger that behavior!
@reissbaker

reissbaker: So I more mean:

1. assistant calls a tool
2. assistant sees the response
3. assistant talks, and then calls another tool
4. assistant sees a response
5. assistant talks, and then calls another tool
6. etc

Here's one prompt to trigger it:

> read some files and tell me about this repo. summarize what you know after each file you read. don't stop until you have a good sense for this repo. make sure not to read any new files until you tell me something first. don't ask permission, just talk to me

image

Since reads and directory lists are permissionless, Octo doesn't wait and prompt for permissions between these calls: it just keeps going.

This is a bit of a contrived example, but I've seen Octo talk in between tool calls organically as well! This was just a simple way to reproduce.

With that sort of behavior + full-height assistant responses, each previous tool call would get quickly pushed offscreen, making it a little annoying to follow along (e.g. Octo might say something in response to a previous tool call, but you can't see what it was since it was pushed offscreen).

sdjidjev:
thanks for your comments @reissbaker!
addressed them with the newest commit, PTAL!
<video src="https://github.com/user-attachments/assets/556e66b6-a8d3-4c0c-8aca-c2a6dcbd089b" />


Tried flinging the scroll wheel a bunch and I don't think I see any leaked scroll events in the Input with this change!
Also this video shows the new autoscroll and psuedo-maxHeight behaviors, definitely dig these suggestions, thank you!

reissbaker (approved):
Amazing! LGTM!

PR #59, opened 2025-12-05
feat(system-prompt): load user-level AGENTS.md

Amolith:
Loads `$XDG_CONFIG_HOME/AGENTS.md` (falling back to `$HOME/.config/AGENTS.md`) using the host filesystem via `process.env` + `fs`, so user-level config applies regardless of Docker container environment.

Also includes a tiny fix for trying to load `~/.config/octofriend/OCTO.md/OCTO.md` instead of `~/.config/octofriend/OCTO.md`; happy to make that a separate PR.

Assisted-by: Claude Opus 4.5 via Crush

### DIFF
--- a/source/system-prompt.ts
+++ b/source/system-prompt.ts
@@ -1,3 +1,4 @@
+import fs from "fs";
 import path from "path";
 import { t, toTypescript } from "structural";
 import { Config } from "./config.ts";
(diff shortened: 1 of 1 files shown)


reissbaker:
Thanks for the contribution! :grin:

PR #60, opened 2025-12-16
Stop flickering when thoughts are shown above content

sdjidjev:
**Issue:** Whenever `reasoningContent` was streamed, any `content` afterwards would flicker, even in a `ScrollView`m caused by re-renders happening off-screen.

**Solution:** Put streamed `thoughts` inside of a `ScrollView`. Also removed `MAX_THOUGHTBOX_WIDTH` as it caused issues in smaller terminals. Once the streaming is complete, show full rendered `thought` text.

 
**Demo of fix:**
<video src="https://github.com/user-attachments/assets/8e19368b-cc22-42b9-be73-6c1f0bc30d46
" />
@reissbaker @billycao

### DIFF
--- a/source/app.tsx
+++ b/source/app.tsx
@@ -2,7 +2,7 @@ import * as fsOld from "fs";
 import React, {
   useState, useCallback, useMemo, useEffect, useRef, createContext, useContext
 } from "react";
-import { Text, Box, Static, measureElement, DOMElement, useInput, useApp, useStdout } from "ink";
+import { Text, Box, Static, measureElement, DOMElement, useInput, useApp } from "ink";
 import clipboardy from "clipboardy";
 import { InputWithHistory } from "./components/input-with-history.tsx";
 import { t } from "structural";
@@ -45,6 +45,7 @@ import { Markdown } from "./markdown/index.tsx";
 import { countLines } from "./str.ts";
 import { VimModeIndicator } from "./components/vim-mode.tsx";
 import { ScrollView, IsScrollableContext } from "./components/scroll-view.tsx";
+import { TerminalSizeTracker, useTerminalSize } from "./components/terminal-size.tsx";
 
 type Props = {
 	config: Config;
@@ -112,19 +113,21 @@ export default function App({ config, configPath, metadata, unchained, transport
         <UnchainedContext.Provider value={unchained}>
           <TransportContext.Provider value={transport}>
             <ExitOnDoubleCtrlC>
-              <Box flexDirection="column" width="100%" height="100%">
-                <Static items={staticItems}>
+              <TerminalSizeTracker>
+                <Box flexDirection="column" width="100%" height="100%">
+                  <Static items={staticItems}>
+                    {
+                      (item, index) => <StaticItemRenderer item={item} key={`static-${index}`} />
+                    }
+                  </Static>
                   {
-                    (item, index) => <StaticItemRenderer item={item} key={`static-${index}`} />
+                    modeData.mode === "responding" &&
+                      (modeData.inflightResponse.reasoningContent || modeData.inflightResponse.content) &&
+                      <MessageDisplay item={modeData.inflightResponse} />
                   }
-                </Static>
-                {
-                  modeData.mode === "responding" &&
-                    (modeData.inflightResponse.reasoningContent || modeData.inflightResponse.content) &&
-                    <MessageDisplay item={modeData.inflightResponse} />
-                }
-                <BottomBar inputHistory={inputHistory} metadata={metadata} />
-              </Box>
+                  <BottomBar inputHistory={inputHistory} metadata={metadata} />
+                </Box>
+              </TerminalSizeTracker>
             </ExitOnDoubleCtrlC>
           </TransportContext.Provider>
         </UnchainedContext.Provider>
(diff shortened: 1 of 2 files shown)


source/app.tsx
```diff
@@ -809,43 +808,48 @@ function AssistantMessageRenderer({ item }: {
...
-              flexShrink={0}
-              width={MAX_THOUGHTBOX_WIDTH - 2}
-              flexDirection="column"
-              marginTop={-1 * Math.max(0, thoughtsOverflow)}
-            >
+      {isStreamingContent ? (
+        <ScrollView height={scrollViewHeight}>
+          {
+            showThoughts && <Box flexDirection="column">
+              <Box
```
sdjidjev: also wouldn't be against removing the inner thoughtbox entirely, would get rid of some weird behavior where if you're not showing the full thoughtbox in the ScrollView it shows blank text
**When whole thoughtbox content is in view**
Screenshot 2025-12-15 at 9 18 41 PM
**When whole thoughtbox content is NOT in view**
Screenshot 2025-12-15 at 9 18 27 PM

source/app.tsx
```diff
@@ -809,43 +808,48 @@ function AssistantMessageRenderer({ item }: {
...
+                  <Text color="gray">{thoughts}</Text>
+                </Box>
+              </Box>
+            </Box>
+          }
+          <Markdown markdown={content} />
+        </ScrollView>
+      ) : (
+        <Box flexDirection="column">
+          {showThoughts &&
```
sdjidjev: we might not even want to show `thoughts` at all once streaming is done, but not totally sure!

reissbaker (approved):
Generally looks good! I was curious about that weird edge case of the thoughts box inside the scrollview not rendering correctly, and poked around the code. Eventually I realized the issue was that `overflow="hidden"` doesn't behave well inside ScrollViews; I changed enough code to figure that out that I figured I'd just push my commit that fixes the issue to this branch.

TLDR of the workaround:

1. I export a new React context from the ScrollView called `IsScrollableContext`, which defaults to `false`. The ScrollView sets up a provider for that context, and sets its state to the boolean that shows whether we've overflowed the window and are scrolling now.
2. In the ThoughtBox, we read that context (it'll be false if we're not inside a ScrollView, or false if we're inside a ScrollView but haven't started scrolling yet).
3. We disable the overflow hidden if we're inside a ScrollView and already started scrolling, since the ScrollView will handle making sure the thoughts don't fill up the entire screen.

I think it's useful to show the thoughts for a couple reasons:

1. Some people like to verify that the model is thinking, and to see why it did whatever it did. It's just helpful from the POV of understanding the agent.
2. We've seen it useful for debugging failed tool calls, where it's clear the model intended to call a tool (it thought about calling it and decided to), and then for some reason failed to then call it.

reissbaker:
Ahh gotcha @sdjidjev !

Just pushed a commit that simplifies the ThoughtBox a little when it's not in the `enforceMaxHeight` state. As far as I can tell (Ghostty on Fedora), it looks like it solves the "thoughtbox intersects the scrollview" issue? Let me know if it works on your machine.

image


AFAICT, the historical borders now only get rendered weirdly if you resize the terminal to be small: if your terminal *starts out* small, it works. Because we're rendering historical messages as StaticItems, we can't redraw them — so redrawing the borders on resize doesn't work. But, they don't look bad for terminals that start out small and stay small: the boxes are drawn given the current terminal window's size, so it should work for small terminals.

I think that's an okay tradeoff, since it's pretty common for terminal tools that don't use alt-mode to get weird on historical resizing. And IMO the box helps make it more clear that these are reasoning tokens, since it's similar to how reasoning gets rendered on ChatGPT (and various riffs of their UX, such as https://synthetic.new :laughing: ).

reissbaker:
You're right: it's actually inconsistent on Fedora+Ghostty, sometimes it works and sometimes it doesn't. Let me play around a little bit...

sdjidjev:
Screenshot 2025-12-16 at 2 34 12 PM

noooo T_T it's still having some weirdness for me. maybe this happens on longer outputs only?
this is the prompt i'm using:
`think a bunch, and then output an example markdown file content. do not create an actual file.`

this feels so close though..... it's almost as if the ScrollView's parent (or grandparent) is changing heights as it's streaming content and that's what causing the issue

reissbaker:
Alright, reproduced the issue when using your prompt, and... it seems fixed now!

Main change was:

1. Actually just read the terminal width directly from stdout
2. Haha, this doesn't work for static items: it always returns `80` when rendering static items. So, when we're not streaming, don't set a width at all.

sdjidjev:
| While streaming | After Streaming |
|--------|-------|
|Screenshot 2025-12-16 at 3 52 52 PM |Screenshot 2025-12-16 at 3 53 18 PM |

dangit i'm still getting issues on small iTerm.... now it seems that the thoughtbox is cutting off content horizontally while streaming (the first screenshot has no right border and the words are cut off).
still getting the repeated lines bug too after streaming is done 😩 

i've noticed as well that two back-to-back runs on normal-sized Terminal will have different spacing above the ScrollView / below the loading text

| Run 1 | Run 2 |
|--------|-------|
| Screenshot 2025-12-16 at 3 58 43 PM | Screenshot 2025-12-16 at 3 58 36 PM |

i think this behavior may be the culprit behind the `After streaming` bug

reissbaker:
@sdjidjev yeee sounds good!

reissbaker:
Alright, think I squashed the final, final edge case with small terminal windows. TLDR:

1. Static items in Ink don't run `useEffect` and don't react to state changes, so our typical way of reading terminal width/height doesn't work when the ThoughtBox is rendered as a static item (since we track state in a ThoughtBox-internal `useState` and then update it with a `useEffect`). This meant that when rendered as a static item, the ThoughtBox always used the default initialized value for the terminal width from the initial `useState` call.
2. It appears there's an internal rendering bug in Ink related to items with borders on small terminal windows, where it'll overflow horizontally non-deterministically depending on the content.
3. So............ we just track terminal size in a `useState` in a React Context, and pull the terminal size from there during rendering instead of reading from stdout. We then manually calculate the correct width for the box. Ink's static items can still correctly pull data from contexts, they just don't update if the data changes (since they're static terminal output).

Annoying, but it seems to work!

reissbaker:
Woooo!

PR #63, opened 2025-12-31
Simplify file tools for more-accurate LLM edits

reissbaker:
Using GLM-4.7 a decent amount w/Octo over the past week, I noticed its reasoning traces were getting pretty confused about the output of our edit tool. It kept thinking the edit somehow got mangled, since we'd output the entire (updated) file after running the tool.

It still eventually did the task I wanted it to, but it obviously wasn't quite at 100% capability, since it kept having to reason about what even had happened.

Poking around Claude Code — presumably the harness GLM trained on, and what most open-source model labs train on at this point — CC doesn't re-output the entire file, or do as much tricky file re-reading as we do (originally, this was intended to make the model less likely to fail to modify a file after the user modified it, since it would always see the latest version; but it's a little confusing honestly for the file to auto-update in the "history" if the model hasn't touched it... Especially if it's modified heavily, and the history of edits stops making sense). TBH, I've long felt the tricky file re-reading was probably a mistake, and even had a TODO to remove it.

This PR cleans up the edit tool output to better match what Claude Code outputs, and removes the tricky file re-reading. Since we no longer do file re-reading in the compilers, it also removes the `transport` argument from the compilers, further simplifying them!

I removed some code duplication in the file read error paths, although there's still a little bit of duplication between the `state.ts` error-transforming code and the `trajectory-arc.ts` error-transforming code. I plan to move the tool execution code out of `state.ts` and into `trajectory-arc.ts` as well, which will allow that last piece of duplication to be excised.

I also fixed a bug introduced by my previous trajectory refactor, where I was mistakenly slicing error IR out that shouldn't have been sliced out.

Testing this with Octo, it feels quite a bit more accurate at editing files and understanding the current state of the code.

### DIFF
--- a/source/compilers/anthropic.ts
+++ b/source/compilers/anthropic.ts
@@ -11,6 +11,7 @@ import { trackTokens } from "../token-tracker.ts";
 import { errorToString } from "../errors.ts";
 import { compactionCompilerExplanation } from "./autocompact.ts";
 import { JsonFixResponse } from "../prompts/autofix-prompts.ts";
+import * as irPrompts from "../prompts/ir-prompts.ts";
 
 const ThinkingBlockSchema = t.subtype({
   type: t.value("thinking"),
@@ -74,20 +75,13 @@ function modelMessageFromIr(
   }
 
   if(ir.role === "file-read") {
-    let content: string;
-    if(seenPath) {
-      content = "File was successfully read.";
-    } else {
-      content = ir.content;
-    }
-
     return {
       role: "user",
       content: [
         {
           type: "tool_result",
           tool_use_id: ir.toolCall.toolCallId,
-          content,
+          content: irPrompts.fileRead(ir.content, seenPath),
         }
       ],
     };
@@ -96,7 +90,7 @@ function modelMessageFromIr(
   if(ir.role === "tool-output" || ir.role === "file-mutate") {
     let content: string;
     if(ir.role === "file-mutate") {
-      content = `${ir.path} was updated successfully.`;
+      content = irPrompts.fileMutation(ir.path);
     } else {
       content = ir.content;
     }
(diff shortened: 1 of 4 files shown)


source/compilers/anthropic.ts
```diff
@@ -21,8 +19,6 @@ const ThinkingBlockSchema = t.subtype({
 });
 
 async function toModelMessage(
```
sdjidjev: neat!! these functions don't need to be `async` anymore now right? (same for all the other compilers)

source/compilers/standard.ts
```diff
@@ -154,45 +150,52 @@ async function llmFromIr(
...
+
   if(ir.role === "tool-reject") {
     return {
       role: "tool",
       tool_call_id: ir.toolCall.toolCallId,
-      content: tagged(TOOL_ERROR_TAG, {}, "Tool call rejected by user. Your tool call did not run."),
+      content: tagged(
+        TOOL_ERROR_TAG,
+        {},
+        "Tool call was rejected by user. Your tool call did not run. No changes were applied."
```
sdjidjev: should we update all the compilers' contents to additionally have `No changes were applied.`? 
i was thinking that it might be nice to have a centralized place for IR content descriptions.
are there situations where we'd want a different explanation of the content between different compilers?

source/compilers/anthropic.ts
reissbaker: True!

reissbaker: Cleaned this up :)

source/compilers/standard.ts
reissbaker: Good idea — will add this to the `prompts` directory. I think in some cases we want to be more/less explicit, depending on the API: for example, we wrap error messages in an `<error>` XML tag to try to be extra convincing to LLMs using the standard API that this was an error; Anthropic, though, has a special `is_error` flag for tool results, so we don't need to waste the tokens for their API. But the inner message should probably be the same!

reissbaker: Done!

PR #65, opened 2026-01-01
Remove windowing and use autocompaction only

reissbaker:
Hot take: I think we should remove IR windowing. Now that autocompaction works, there's no need for it! Here are some downsides to windowing:

1. It prevents prompt caching. Since the message list becomes a rolling slice of the history, the first message constantly changes as new messages are added to the end of the list — which ends up removing the ability for backends to prompt cache. For pay-per-token API providers, this increases the cost for our users (since most usage-based APIs have discounts for prompt caching); for subscription services (like Synthetic!), this slows down time-to-first-token for our users since we have to recompute the entire KV cache.
2. When using windowing *and* autocompaction, in theory the window should never apply: the model should always autocompact before it hits the context threshold used for windowing. Arguably it *could* apply if the compaction checkpoint still is too large to fit, but then... the agent would just lose the entire conversation history. I think there's not much value to windowing anymore.

This PR removes IR windowing entirely, and also removes the ability to turn off autocompaction (since without IR windowing, autocompaction is a necessary feature).

I think it would be nice for our 0.0.49 release to just hard cut over to autocompaction and remove windowing. That way we won't have to maintain the config option, and we won't need to nag people in the future if/when we drop IR windowing about having an outdated config. Let's just ship autocompaction!

### DIFF
--- a/source/compilers/standard.ts
+++ b/source/compilers/standard.ts
@@ -294,7 +294,7 @@ export const runAgent: Compiler = async ({
   const curl = generateCurlFrom({
     baseURL: model.baseUrl,
     model: model.model,
-    messages: messages,
+    messages,
     ...toolsParam,
   });
   return await handleKnownErrors(curl, async () => {


source/compilers/standard.ts
```diff
@@ -297,7 +294,7 @@ export const runAgent: Compiler = async ({
   const curl = generateCurlFrom({
     baseURL: model.baseUrl,
     model: model.model,
-    messages,
+    messages: messages,
```
sdjidjev: -super nit- you can just leave this as `messages,`

reissbaker: derp good call

PR #66, opened 2026-01-01
Let our lil cephalopod buddy learn (Agent) skills

Amolith:
Discover SKILL.md files from configurable paths, parse YAML frontmatter for metadata, validate structure, and inject available skills into the system prompt as XML for LLM context.

Includes symlink-aware directory traversal and duplicate name detection.

Assisted-by: Claude Opus 4.5 via Crush

### DIFF
--- a/source/config.ts
+++ b/source/config.ts
@@ -59,7 +59,9 @@ const ConfigSchema = t.exact({
   defaultApiKeyOverrides: t.optional(t.dict(t.str)),
   mcpServers: t.optional(t.dict(McpServerConfigSchema)),
   autoCompact: t.optional(AutoCompactConfigSchema),
-  skillsPaths: t.optional(t.array(t.str)),
+  skills: t.optional(t.exact({
+    paths: t.optional(t.array(t.str)),
+  })),
 });
 export type Config = t.GetType<typeof ConfigSchema>;
 export const AUTOFIX_KEYS = [
--- a/source/prompts/system-prompt.ts
+++ b/source/prompts/system-prompt.ts
@@ -212,8 +212,8 @@ ${mcpSections.join('\n\n')}
 async function skillsPrompt(transport: Transport, signal: AbortSignal, config: Config) {
   const skillsPaths: string[] = [];
 
-  if (config.skillsPaths && config.skillsPaths.length > 0) {
-    skillsPaths.push(...config.skillsPaths);
+  if (config.skills?.paths && config.skills.paths.length > 0) {
+    skillsPaths.push(...config.skills.paths);
   } else {
     const defaultPath = await getDefaultSkillsPath(transport, signal);
     skillsPaths.push(defaultPath);


source/skills/skills.ts
```diff
@@ -0,0 +1,219 @@
...
+
+        if (seenNames.has(skill.name)) {
+          console.warn(`Duplicate skill name "${skill.name}" at ${filePath}, skipping`);
+          return;
+        }
+        seenNames.add(skill.name);
+
+        skills.push(skill);
+      } catch (e) {
+        console.warn(`Error reading skill file ${filePath}: ${e}`);
```
reissbaker: Minor nit: prefer

```
import * as logger from "../logger.ts";

logger.error("info", "...");
```

source/config.ts
```diff
@@ -59,6 +59,7 @@ const ConfigSchema = t.exact({
   defaultApiKeyOverrides: t.optional(t.dict(t.str)),
   mcpServers: t.optional(t.dict(McpServerConfigSchema)),
   autoCompact: t.optional(AutoCompactConfigSchema),
+  skillsPaths: t.optional(t.array(t.str)),
```
reissbaker: Config suggestion: rather than `skillsPaths` as a top-level string, I'd suggest:

```typescript
skills: t.optional(t.exact({
  paths: t.optional(t.array(t.str)),
})),
```

I suspect we may want more skills-related configs in the future, so this allows us to do that cleanly without more top-level keys.

source/skills/skills.ts
```diff
@@ -0,0 +1,219 @@
...
+    path: path.dirname(filePath),
+    skillFilePath: filePath,
+  };
+}
+
+async function walkDirectory(
+  transport: Transport,
+  signal: AbortSignal,
+  dirPath: string,
+  callback: (filePath: string) => Promise<void>
```
reissbaker: Minor style nit: I think this pattern is most idiomatically written as an async generator (`async function*` rather than `async function`), since there's only a single callback and the function otherwise doesn't return anything.

```typescript
async function* walkDirectory(
  transport: Transport,
  signal: AbortSignal,
  dirPath: string,
): AsyncGenerator<string> {
  // code...

  yield fullPath; // instead of callback(fullPath)

  // code...
}
```

That way you can easily use it in an async for loop:

```typescript
for await(const filePath of walkDirectory(...)) {
  if(signal.aborted) return;
  if(seen.has(filePath)) continue;
  // ... code
}
```

reissbaker (commented):
Code structure generally looks good to me! This is great, thank you so much — I was literally about to build this feature myself. Left a couple code style comments, once those are done LGTM and I will merge :)

I'm back and forth on whether it's worth writing a specialized tool for skills rather than just instructing Octo to use the existing `read` tool and injecting the available skills into the system prompt; there are pros and cons. Let's go with your simple approach for now and I might add a specialized tool later.

source/skills/skills.ts
Amolith: Fixed in 5de9124ff98dc512a479f26f9ee5dc655b8e70f3

source/config.ts
Amolith: Fixed in 5d2da73c52e7cdf0c6bd7854210b39063243cbd3

source/skills/skills.ts
Amolith: Fixed in b4c4fcbb35b74377e0f6dfb7d3ad3db5161a7c27

PR #68, opened 2026-01-06
Toggle unchained mode with shift+tab

sdjidjev:
Screenshots:

| Launched with unchained mode | Shift tabbed to unchained mode (from Review mode) | Review Mode |
| :--- | :--- | :--- |
| Screenshot 2026-01-05 at 5 04 18 PM | Screenshot 2026-01-05 at 5 38 58 PM | Screenshot 2026-01-05 at 5 39 13 PM |

@reissbaker @billycao

### DIFF
--- a/source/app.tsx
+++ b/source/app.tsx
@@ -120,9 +120,9 @@ export default function App({
     setIsUnchained(prev => {
       const unchained = !prev;
       if(unchained) {
-        setTempNotification(<><Text bold>⚡ UNCHAINED MODE:</Text> Octo runs edits and shell commands automatically</>);
+        setTempNotification(<>Octo runs edits and shell commands automatically</>);
       } else {
-        setTempNotification(<><Text bold>Collaboration mode:</Text> Octo asks permission before running edits or shell commands</>);
+        setTempNotification(<>Octo asks permission before running edits or shell commands</>);
       }
       return unchained;
     });
@@ -214,7 +214,7 @@ function BottomBar({ inputHistory, metadata, tempNotification }: {
   const unchained = useUnchained();
 
   return <Box flexDirection="column" width="100%">
-    <BottomBarContent inputHistory={inputHistory} tempNotification={displayedTempNotification} />
+    <BottomBarContent inputHistory={inputHistory} />
     <Box
       width="100%"
       justifyContent="space-between"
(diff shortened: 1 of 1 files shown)


source/app.tsx
```diff
@@ -174,9 +182,16 @@ function BottomBar({ inputHistory, metadata }: {
...
-      <Text color={themeColor}>
-        { ctrlCPressed && "Press Ctrl+C again to exit." }
-      </Text>
+      <Box>
+        <Text color={themeColor}>
+          { ctrlCPressed && "Press Ctrl+C again to exit." }
+        </Text>
+        { !ctrlCPressed && (
+          <Text color={unchained ? themeColor : "gray"}>
+            { unchained ? "⚡ UNCHAINED MODE" : "Review mode" } <Text dimColor>(Shift+Tab to toggle)</Text>
```
reissbaker: Nit: it might be useful to offer some explanatory text. For example:

```
UNCHAINED MODE: Octo will not require permission before editing or using tools
Review mode: Octo will ask for permission before editing or using shell commands
```

I imagine you didn't do this because it's wordy af. One workaround could be to only show it for the first ~15 seconds or so, using a `setState` + `useEffect` + `setTimeout`

reissbaker: Looking at the screenshot, I imagine the update notification as is might conflict with the long wordy message: maybe we swap the location of the "ESC to enter the menu" and the update notification, to make room for both?

reissbaker: Hmmm it might still be too long TBH. Maybe we print it out as `type: "notification"` text? (`Octo is in Review mode and will ask for permission before editing or using shell commands`)

reissbaker: Also minor nit but maybe `Collaboration` mode is a clearer name? `Review` mode sounds like *Octo* might be doing reviews of something, whereas `Collaboration` mode makes it more clear IMO that Octo is expecting to work with you as opposed to working by itself.

reissbaker:
Hmmm. I still think it looks a little weird to have the mode mentioned in two places in the UI, and on larger terminals since it's so long / sentence-like, it feels off-center since there's a sentence right above it that's left-aligned and this one is right-aligned. I think combining both at the bottom makes more sense / looks better. We can fade out the explanation after a few seconds though. Here's a random sketch I just did — not sure whether it looks better to have it multiline or one single line, but on small screens maybe one line would wrap weirdly so multiline is better?

octo-sketch

The first sketch is boot, and the second sketch is a few seconds after boot when the explanation fades out.

source/app.tsx
```diff
@@ -153,11 +168,15 @@ export default function App({
   </SetConfigContext.Provider>
 }
 
-function BottomBar({ inputHistory, metadata }: {
+function BottomBar({ inputHistory, metadata, tempNotification }: {
   inputHistory: InputHistory
   metadata: Metadata,
+  tempNotification: React.ReactNode | null,
```
sdjidjev, line 174: kept this as a node even though it's not needed for the mode toggle in case we want fancier formatting later on....

PR #72, opened 2026-01-09
React+Ink upgrade for React compiler

reissbaker:
This upgrades our React+Ink library versions so that we can use the [React Compiler](https://react.dev/learn/react-compiler)! The React compiler auto-optimizes our React components, making sure e.g. callbacks are memoized automatically.

The React Compiler requires a build pipeline. The simplest path forward was just to use Babel (the compiler itself is a Babel plugin, and using Vite required a bunch of boilerplate to get the plugin working), so I added Babel and updated the build command to use it.

Upgrading React+Ink made an occasional bug that I noticed — diff padding collapse — an *always* bug rather than an occasional one. I believe this is in fact a bug in our code, and Ink was previously rendering incorrectly: whitespace *should* collapse unless you're setting an explicit width, and we were using whitespace to indent rather than using `padding` or setting widths. For example, here's how Octo post-update would render a diff of `loading.tsx`:

```typescript
const DEFAULT_LOADING_STRINGS = [
"Scheming",
"Plotting",
// etc
```

Notice that the whitespace has collapsed and no longer appears indented. Here's how it should render:

```typescript
const DEFAULT_LOADING_STRINGS = [
  "Scheming",
  "Plotting",
  // etc
```

To fix this I just added `paddingLeft` instead of whitespace, and the diffs now consistently render correctly. I'm not sure why this ever worked in previous versions of Ink, but it did!

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -27,6 +27,7 @@
 		"noUncheckedSideEffectImports": true,
 		"forceConsistentCasingInFileNames": true,
     "rewriteRelativeImportExtensions": true,
-    "allowImportingTsExtensions": true
+    "allowImportingTsExtensions": true,
+    "noEmit": true
 	}
 }


sdjidjev (approved):
interesting that it only sometimes collapsed the whitespace characters before 🤔 
seems better this way anyways!

PR #75, opened 2026-01-09
Instructions that show up on Windows execution (when not using bash)

sdjidjev:
On first time setup only, we check if user is on Windows. If they are, and bash is not available (tested by running `bash --version`), then we prompt them to install either Git Bash or WSL and run it on there. (put Git Bash ahead of WSL because the local file system is easily accessible with the former)

**Screenshots:**
Loading (only on Windows environment during first time setup, only takes about a second)
Screenshot 2026-01-08 231828

Installation instructions:
Screenshot 2026-01-08 232603

On Git Bash in Windows environment:
Screenshot 2026-01-08 232319

### DIFF
--- a/source/first-time-setup.tsx
+++ b/source/first-time-setup.tsx
@@ -3,6 +3,7 @@ import { Box, Text, useInput, useApp } from "ink";
 import fs from "fs/promises";
 import path from "path";
 import json5 from "json5";
+import { execSync } from "child_process";
 import TextInput from "./components/text-input.tsx";
 import { Config } from "./config.ts";
 import { useColor } from "./theme.ts";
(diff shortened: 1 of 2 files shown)


source/first-time-setup.tsx
```diff
@@ -346,11 +357,59 @@ function AutofixCompleteScreen({ onContinue }: { onContinue: () => void }) {
   </CenteredBox>
 }
 
+function WindowsSetupInstructions({ bashAvailable }: { bashAvailable: boolean | null }) {
+  const app = useApp();
+
+  useLayoutEffect(() => {
+    if(bashAvailable === false) {
+      app.exit();
```
sdjidjev, line 365: since the shell isn't compatible, i think it's better to exit than to leave the thread running

PR #76, opened 2026-01-09
Text handling improvements

reissbaker:
Previously, typing a super long prompt would result in some annoying flicker on keystrokes, because React would re-render the entire prompt, since it's all a single `<Text>` component.

This wraps each line in its own `<Box>`, so that React sees that only a single component in the tree changed (the line you're currently editing) rather than needing to re-render the entire `<Text>` component when only a single character changed.

This also somewhat lays groundwork for future typing improvements, e.g. `dd` in Vim only deleting the line you're actually on vs deleting to the last explicit newline; `ctrl-a` and `ctrl-e` correctly handling multiline text; etc.

### DIFF
--- a/source/components/text-input.tsx
+++ b/source/components/text-input.tsx
@@ -32,36 +32,36 @@ export default function TextInput({
 	vimMode = 'NORMAL',
 	setVimMode,
 }: Props) {
-	const [state, setState] = useState({
-		cursorOffset: 0,
-		cursorWidth: 0,
-	});
-	const [isInitializing, setIsInitializing] = useState(true);
-	const [measuredWidth, setMeasuredWidth] = useState(0);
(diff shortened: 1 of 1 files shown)


source/components/text-input.tsx
```diff
@@ -48,6 +52,28 @@ export default function TextInput({
 		return () => clearTimeout(timer);
 	}, []);
 
+  function handleElementSize() {
+		if(containerRef.current) {
```
sdjidjev: -super nit- indentation
i see some inconsistent spacing throughout this file idk if github is weirding out or not

reissbaker: Whoof — I think this is probably mixed tabs/spaces from many people's editors + various LLMs. I just de-dented and re-indented most of the file, but probably we should just set up Prettier and autoformat the codebase.

PR #77, opened 2026-01-10
UI tweaks for chained/unchained toggling

reissbaker:
Couple tiny UI tweaks after playing around with the (awesome!) toggle.

1. Always reserves space for the temp notification, even if it's not currently showing, to prevent the UI from jumping around when you're toggling modes — otherwise, the UI jumps upwards when you toggle, since previously the temp notification took up 0 space.
2. Slightly toned down the unchained mode text: it already turns the whole UI red so it felt like we didn't need to capitalize it and make the (non-temporary) mode text red too.
3. Shows the notif about modes on boot, since most people won't know what "Collaboration mode" means at first.

Ignore the fact that my terminal font doesn't have a nice emoji for :zap: ! It's still the :zap: emoji, I didn't change that.

With temp notif displayed:
image

Without:
image

### DIFF
--- a/source/app.tsx
+++ b/source/app.tsx
@@ -238,7 +238,7 @@ function BottomBar({ inputHistory, metadata, tempNotification }: {
       </Box>
       <Text color={themeColor}>{versionCheck}</Text>
     </Box>
-    <Box height={1}>
+    <Box minHeight={1}>
       {displayedTempNotification && (
         <Box width="100%" flexShrink={0}>
           <Text color={themeColor} wrap="wrap">{displayedTempNotification}</Text>


source/app.tsx
```diff
@@ -222,23 +224,25 @@ function BottomBar({ inputHistory, metadata, tempNotification }: {
...
         )}
       </Box>
       <Text color={themeColor}>{versionCheck}</Text>
     </Box>
-    {displayedTempNotification && (
-      <Box width="100%" flexShrink={0}>
-        <Text color={themeColor} wrap="wrap">{displayedTempNotification}</Text>
-      </Box>
-    )}
+    <Box height={1}>
```
sdjidjev: would this mean it doesn't wrap btw? it's ok if not, just checking! we can remove that prop in there if so

reissbaker: Fixed! Good catch, I swapped this to minHeight

PR #79, opened 2026-01-13
Set up whitelist system for tool requests to auto-allow

sdjidjev:
### Major Changes:
~~1. **Moved the tool definition files into categories folders**~~

~~2. For all shell commands, we take at most the first two segments of the command~~
3. File operations are whitelisted to be allowed for the entire directory that `octo` is in.
- i think it would be too finicky to have to manually whitelist file operations for specific directories within
**Screenshots:**
Screenshot 2026-01-12 at 7 27 49 PM

### DIFF
--- a/source/hooks/use-cwd.tsx
+++ b/source/hooks/use-cwd.tsx
@@ -1,0 +1,8 @@
+import { createContext, useContext } from "react";
+
+const nullByte = "\0";
+export const CwdContext = createContext<string>(nullByte);
+
+export function useCwd(): string {
+  return useContext(CwdContext);
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -252,6 +252,7 @@
       "integrity": "sha512-e7jT4DxYvIDLk1ZHmU/m/mB19rex9sv0c2ftBtjSBv+kVM/902eh0fINUzD7UwLLNR+jU585GxUJ8/EBfAM5fw==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.5",
@@ -1796,6 +1797,7 @@
       "integrity": "sha512-NMv9ASNARoKksWtsq/SHakpYAYnhBrQgGD8zkLYk/jaK8jUGn08CfEdTRgYhMypUQAfzSP8W6gNLe0q19/t4VA==",
       "devOptional": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@types/node": "*"
       }
@@ -1861,6 +1863,7 @@
       "integrity": "sha512-MWtvHrGZLFttgeEj28VXHxpmwYbor/ATPYbBfSFZEIRK0ecCFLl2Qo55z52Hss+UV9CRN7trSeq1zbgx7YDWWg==",
       "devOptional": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "csstype": "^3.2.2"
       }
@@ -2253,6 +2256,7 @@
       "integrity": "sha512-eGbYq2CT+tos1fBwLQ/tkBt9J5M3JEHjku4hbvQUePCckkvVf14xWj+1m7dGoK81M/fOjFT7yM9UMeKT/+vFLQ==",
       "hasInstallScript": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "bindings": "^1.5.0",
         "prebuild-install": "^7.1.1"
@@ -2363,6 +2367,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "baseline-browser-mapping": "^2.9.0",
         "caniuse-lite": "^1.0.30001759",
@@ -2661,6 +2666,7 @@
       "resolved": "https://registry.npmjs.org/commander/-/commander-13.1.0.tgz",
       "integrity": "sha512-/rFeCpNJQbhSZjGVwO9RFV3xPqbnERS8MmIQzCtD/zl6gpJuV/bMLuN92oG3F7d8oDEHHRrujSXNUr8fpjntKw==",
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=18"
       }
@@ -3174,6 +3180,7 @@
       "dev": true,
       "hasInstallScript": true,
       "license": "MIT",
+      "peer": true,
       "bin": {
         "esbuild": "bin/esbuild"
       },
@@ -3355,6 +3362,7 @@
       "resolved": "https://registry.npmjs.org/express/-/express-5.2.1.tgz",
       "integrity": "sha512-hIS4idWWai69NezIdRt2xFVofaF4j+6INOpJlVOLDO8zXGpUVEVzIYk12UUi2JzjEzWL3IOAxcTubgz9Po0yXw==",
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "accepts": "^2.0.0",
         "body-parser": "^2.2.1",
@@ -3722,6 +3730,19 @@
         "node": ">= 6"
       }
     },
+    "node_modules/glob/node_modules/minimatch": {
+      "version": "3.1.2",
+      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-3.1.2.tgz",
+      "integrity": "sha512-J7p63hRiAjw1NDEww1W7i37+ByIrOWO5XQQAzZ3VOcL0PNybwpfmV/N05zFAzwQ9USyEcX6t3UO+K5aqBQOIHw==",
+      "dev": true,
+      "license": "ISC",
+      "dependencies": {
+        "brace-expansion": "^1.1.7"
+      },
+      "engines": {
+        "node": "*"
+      }
+    },
     "node_modules/gopd": {
       "version": "1.2.0",
       "resolved": "https://registry.npmjs.org/gopd/-/gopd-1.2.0.tgz",
@@ -3768,9 +3789,9 @@
       }
     },
     "node_modules/hono": {
-      "version": "4.11.7",
-      "resolved": "https://registry.npmjs.org/hono/-/hono-4.11.7.tgz",
-      "integrity": "sha512-l7qMiNee7t82bH3SeyUCt9UF15EVmaBvsppY2zQtrbIhl/yzBTny+YUxsVjSjQ6gaqaeVtZmGocom8TzBlA4Yw==",
+      "version": "4.11.4",
+      "resolved": "https://registry.npmjs.org/hono/-/hono-4.11.4.tgz",
+      "integrity": "sha512-U7tt8JsyrxSRKspfhtLET79pU8K+tInj5QZXs1jSugO1Vq5dFj3kmZsRldo29mTBfcjDRVRXrEZ6LS63Cog9ZA==",
       "license": "MIT",
       "engines": {
         "node": ">=16.9.0"
@@ -4611,19 +4632,6 @@
         "url": "https://github.com/sponsors/sindresorhus"
       }
     },
-    "node_modules/minimatch": {
-      "version": "3.1.2",
-      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-3.1.2.tgz",
-      "integrity": "sha512-J7p63hRiAjw1NDEww1W7i37+ByIrOWO5XQQAzZ3VOcL0PNybwpfmV/N05zFAzwQ9USyEcX6t3UO+K5aqBQOIHw==",
-      "dev": true,
-      "license": "ISC",
-      "dependencies": {
-        "brace-expansion": "^1.1.7"
-      },
-      "engines": {
-        "node": "*"
-      }
-    },
     "node_modules/minimist": {
       "version": "1.2.8",
       "resolved": "https://registry.npmjs.org/minimist/-/minimist-1.2.8.tgz",
@@ -4946,6 +4954,7 @@
       "integrity": "sha512-5gTmgEY/sqK6gFXLIsQNH19lWb4ebPDLA4SdLP7dsWkIXHWlG66oPuVvXSGFPppYZz8ZDZq0dYYrbHfBCVUb1Q==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=12"
       },
@@ -5153,6 +5162,7 @@
       "resolved": "https://registry.npmjs.org/react/-/react-19.2.3.tgz",
       "integrity": "sha512-Ku/hhYbVjOQnXDZFv2+RibmLFGwFdeeKHFcOTlrt7xplBnya5OGn/hIRDsqDiSUcfORsDC7MPxwork8jBwsIWA==",
       "license": "MIT",
+      "peer": true,
       "engines": {
         "node": ">=0.10.0"
       }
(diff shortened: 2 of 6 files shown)


source/compilers/autofix.ts
```diff
@@ -1,12 +1,9 @@
 import OpenAI from "openai";
 import { t } from "structural";
 import { Config, assertKeyForModel } from "../config.ts";
-import { ArgumentsSchema as EditSchema } from "../tools/tool-defs/edit.ts";
```
reissbaker, line 4: Nit: `kebab-case` for files/directories!

source/tools/permissions/tool-pattern-extractor.ts
```diff
@@ -0,0 +1,97 @@
+import { ToolCallRequest } from "../../ir/llm-ir.ts";
+import { WhitelistType } from "./index.ts";
+import {
+  extractCommandPrefix,
+  extractMcpPattern,
+} from "./whitelist.ts";
+import { cwd } from "process";
+import { TOOL_CATEGORIES } from "../tool-defs/categories.ts";
+
+export interface ToolPermissionInfo {
```
reissbaker: Nit: generally prefer `type X = { ... }` to `interface X { ... }` unless you're intending for a class to later implement the interface

source/tools/permissions/tool-pattern-extractor.ts
```diff
@@ -0,0 +1,97 @@
...
+  type: WhitelistType;
+  value: string;
+  pattern: string;
+  label: string;
+  labelParts: { text: string; bold?: boolean }[];
+}
+
+interface ToolConfig {
+  type: WhitelistType;
+  extractValue: (args: Record<string, unknown>) => string;
```
reissbaker: I think this `unknown` is biting you later and forcing a lot of `as string`. I think you could rewrite this to:

```typescript
type ToolConfig<CategoryType> {
  extractValue: (args: Record<string, CategoryType>) => string
}
```

That being said, I don't know that you necessarily want to type every return value as a string — especially since you then need to do string parsing later with comparisons to extracted patterns (and I also don't think this will reliably work for shell command whitelisting! E.g. consider `cp x y && rm z`: the user might have whitelisted `cp` but not have expected `rm` to be similarly whitelisted).

I generally like the idea of categories, but I think each category probably defines its own whitelisting scheme. For example:

```typescript
// in some common file:
type AddToWhitelist<Input, Whitelist> = (input: Input, whitelist: Whitelist) => Promise<Whitelist>;
type IsWhitelisted<Input, Whitelist> = (input: Input, whitelist: Whitelist) => Promise<boolean>;

// in file-operations/index.ts:
type FilePath = string;
export type FileWhitelist = { directories: string[] };

export const addToWhitelist: AddToWhitelist<FilePath, FileWhitelist> = async (filepath, whitelist) => {
  return {
    directories: [
      ...whitelist.directories,
      path.dirname(filepath),
    ]
  };
};
export const isWhitelisted: IsWhitelisted<FilePath, FileWhitelist> = async (filepath, whitelist) => {
  for(const dir of whitelist.directories) {
    if(filepath.startsWith(dir)) return true; // or use minimatch
  }
  return false;
};
```

Then you'd have some larger function somewhere that operates on all categories, and returns an overall merged whitelist i.e. `{ fileOperations: FileWhitelist, ... }` that you store and manage in the Zustand state.

Generally speaking you want these to be async: the way e.g. Claude Code handles Bash whitelisting is by constructing a prompt to itself that just says something like:

```markdown
The user has whitelisted the following shell commands:
{ insert whitelist }

The system is trying to determine if the following shell command is in the whitelist:
{ insert shell command }
Call the Allow tool to allow this command,  or the Disallow tool if the command is different than the whitelisted ones. Check for shell escapes or tricky usage of Bash like &&, ||, etc: only commands that are materially the same should pass the whitelist.
```

IMO your initial PR can leave this part out and just always `return false` for the `isWhitelisted` check for shell commands for now!

sdjidjev: Oh yeah, great catch with the `&&` shell commands, definitely would complicate things. 🤔 
But wow, Claude Code's solution for it makes a lot of sense lol, sure beats wild regex strings!

And good call in putting the category's scheme per folder, definitely better than trying to store all the different ones in this file.

source/tools/permissions/tool-permission-info.ts
```diff
@@ -0,0 +1,86 @@
...
+export function getToolOperationArgs<T extends WhitelistCategory>(
+  toolCategory: T,
+  toolReq: ToolCallRequest,
+  transport: Transport,
+): AnyToolArgs {
+  switch (toolCategory) {
+    case "fileOperations":
+      return { transport, abortSignal: new AbortController().signal } as FileOperationArgs;
+    case "command":
+      return toolReq.function.arguments as CommandArgs;
```
reissbaker: These typecasts are a little scary / footgun-y to me. The problem here is that you don't know what the type of `toolReq` is, and you can't infer it from the category since there's no type-level mapping in the code of category -> tool, so you're forced into typecasting here.

I think it's probably worth not having this function, and creating the args at the place they're referenced, since that codepath knows what type `toolReq` is. On line 648 of `app.tsx`, there's a large switch statement that checks every possible tool type and returns different stuff based on the tool type. Currently, it only returns a component; however, I think you could return something like:

```
{
  operationArgs: ...,
  component: ...
}
```

Which would be type-safe and ensure that we never forget to add handling here. As far as I can tell that's the only component that calls `getToolOperationArgs`, so you could then delete this function!

source/tools/permissions/tool-permission-info.ts
```diff
@@ -0,0 +1,86 @@
...
+type AnyToolArgs = CategoryArgsMap[keyof CategoryArgsMap];
+
+export function getToolOperationArgs<T extends WhitelistCategory>(
+  toolCategory: T,
+  toolReq: ToolCallRequest,
+  transport: Transport,
+): AnyToolArgs {
+  switch (toolCategory) {
+    case "fileOperations":
+      return { transport, abortSignal: new AbortController().signal } as FileOperationArgs;
```
reissbaker: Huh. This looks like a bug to me: if the code wants an abortSignal, presumably it's an operation that should be abortable. But this code creates and immediately discards a new `AbortController`, which means the user will never be able to abort it: only the `AbortController` can abort its signal, but the reference to the controller is immediately lost.

source/tools/permissions/merged-whitelist.ts
```diff
@@ -0,0 +1,91 @@
+import * as fileOperationsWhitelist from "../tool-defs/file-operations/index.ts";
```
reissbaker: Nit: instead of having two files, `merged-whitelist.ts` that operates on whitelist types, and a second file that wraps `merged-whitelist.ts` with different function names, I think you should probably just have one file with everything called `category-whitelist.ts`, and where the functions are named the way you want to export them.

source/tools/permissions/merged-whitelist.ts
```diff
@@ -0,0 +1,91 @@
...
+import * as commandWhitelist from "../tool-defs/command/index.ts";
+import * as mcpWhitelist from "../tool-defs/mcp/index.ts";
+import * as fetchWhitelist from "../tool-defs/fetch/index.ts";
+import * as skillWhitelist from "../tool-defs/skill/index.ts";
+
+export interface MergedWhitelist {
```
reissbaker: Nit: don't use `interface` unless you're planning on making classes that implement the interface: instead use `type MergedWhitelist = { ... }`

source/tools/permissions/tool-permission-info.ts
```diff
@@ -0,0 +1,86 @@
...
+  category: WhitelistCategory;
+  getPermissionWhitelistKey: (toolName: T, args: ToolArgs<T>) => string;
+  formatLabelParts: (key: string, context: LabelContext) => { text: string; bold?: boolean }[];
+};
+
+type ToolConfigs = {
+  [T in AllToolsFromDefs]: ToolConfig<T>;
+};
+
+type CategoryConfigsType = {
```
reissbaker: Unused code?

source/tools/permissions/tool-permission-info.ts
```diff
@@ -0,0 +1,86 @@
...
+  return config.getPermissionWhitelistKey(toolName, args);
+}
+
+type ToolConfig<T extends AllToolsFromDefs> = {
+  category: WhitelistCategory;
+  getPermissionWhitelistKey: (toolName: T, args: ToolArgs<T>) => string;
+  formatLabelParts: (key: string, context: LabelContext) => { text: string; bold?: boolean }[];
+};
+
+type ToolConfigs = {
```
reissbaker: Unused code?

source/tools/permissions/tool-permission-info.ts
```diff
@@ -0,0 +1,86 @@
...
+  args: CategoryArgsMap[T],
+): Promise<string> {
+  const config = categoryConfigs[category] as CategoryConfig<
+    WhitelistCategoryData,
+    CategoryArgsMap[T]
+  >;
+  return config.getPermissionWhitelistKey(toolName, args);
+}
+
+type ToolConfig<T extends AllToolsFromDefs> = {
```
reissbaker: Also seems unused (outside of other types that are also unused)

source/tools/permissions/category-whitelist-types.ts
```diff
@@ -0,0 +1,39 @@
...
+  | SkillArgs
+  | WebSearchArgs;
+
+export type CategoryConfig<
+  TWhitelist extends WhitelistCategoryData,
+  TArgs extends ToolOperationArgs,
+> = {
+  getPermissionWhitelistKey: (toolName: string, args: TArgs) => Promise<string>;
+  getPermissionContext?: (args: TArgs) => Promise<string>;
+  yesAndAlwaysAllowLabelSuffix: (
```
reissbaker: This function along with `getPermissionContext` (which is only called to fetch data to input into `yesAndAlwaysAllowLabelSuffix`) both feel like UI code: IMO, those should probably live in the React components rather than living in the tool definitions.

There's an especially awkward part of them living in the tool definitions: in *most* cases, these are synchronous and can be called without worrying about loading states or aborts, etc. However, for file operations, they aren't since we check the `cwd`, which may be async! So, we would potentially have to worry about loading states and aborts.

I think you can solve this differently. We know the `cwd` never changes after boot, so we can load it in `cli.tsx` and pass it in as a property to the App. The App can then thread that down to the various React components, either as a property or using React Context. Then, we can render the correct label in the React components synchronously: the `await` was already handled and we know the CWD isn't going to have changed since then. Personally, I like React Context for this, since actually I can imagine a future in which `cwd` could change at specific points: if we allow historical session loading, we'd probably want Octo to use the previous session's `cwd` or else it'll get pretty messed up (i.e. could load the wrong OCTO.md, etc). But we'd only need to handle the loading state in one place: the place where we load the previous session. After that, `cwd` becomes static again, and we should be able to access it without doing `async` stuff.

Moving this code into the React component layer also means you don't need to create new `<Text>`-like types to track whether certain substrings should be bold or not: instead, you can use regular `<Text>` components!

Line 648 of `app.tsx` already has a switch statement for different tools that handle all the different options for different types. You can probably do something similar. My hunch is that the switch statement is getting pretty big, and you'll probably want to break those out into special per-tool renderers, or perhaps per-category renderers that take a specific subset of tools as props (i.e. the FileOperationsPermission` only takes the specific file tools we have defined as props, and thus knows that it can safely access paths, etc from the arguments).

In general, I think one nice mental model for Octo is: it's actually two things. There's:

1. `libocto`, which is the basics of tool calls and running trajectories.
2. `octo`, which is the terminal UI that uses `libocto`.

In the future, we may want alternative UIs beyond just our current terminal-focused one (a browser-based one?). And we may want to use the `libocto` harness to power non-Octo experiences, like allowing web search / fetch tools on Synthetic. So, I think it's worth keeping Octo UI code outside of `libocto`. But even if that weren't the case, I also think it's just more maintainable that way!

source/tools/permissions/merged-whitelist.ts
```diff
@@ -0,0 +1,91 @@
...
+      return await skillWhitelist.isWhitelisted();
+  }
+}
+
+export const categoryConfigs = {
+  fileOperations: fileOperationsWhitelist.config,
+  command: commandWhitelist.config,
+  mcp: mcpWhitelist.config,
+  fetch: fetchWhitelist.config,
+  skill: skillWhitelist.config,
```
reissbaker: If you remove the `getPermissionContext` and `yesAndAlwaysAllowLabelSuffix` functions as suggested above, you actually don't need to export special `config` objects from the category-specific files: instead, you could do:

```typescript
import * as fileOperationsWhitelist;
// ...

export const categoryWhitelists = {
  // since the module defines addToWhitelist and isWhitelisted directly as module functions,
  // you don't need an extra object: the module itself satisfies the type
  fileOperations: fileOperationsWhitelist,
  // ...
};
```

This would also remove the `config` name, which I think helps make the code more greppable: we already use the term `config` to refer to the Octofriend config. This new code isn't defining a config per se, it's a set of whitelisting operations on categories (once the other two fns are removed!).

source/tools/permissions/tool-permission-info.ts
```diff
@@ -0,0 +1,86 @@
...
+  >;
+  return config.getPermissionContext?.(args) ?? "";
+}
+
+export async function getPermissionWhitelistKey<T extends WhitelistCategory>(
+  category: T,
+  toolName: string,
+  args: CategoryArgsMap[T],
+): Promise<string> {
+  const config = categoryConfigs[category] as CategoryConfig<
```
reissbaker: I think you could get rid of this `as` by calling the underlying function in the React layer, i.e in a file read component doing `categoryConfigs.fileOperations(tool.name, tool.arguments)`: I think this wrapper function doesn't save a lot of lines of code compared to that, and introduces a layer that hides the underlying types from the TS compiler. I'd probably remove this function.

reissbaker (changes requested):
Sorry for the slow re-review!

I think this is a lot closer. I recommended moving a few pieces to the React layer, which I think will help make the code cleaner and also remove some typecasting you're doing.

Re: ensuring all tools have a permissions category defined, I think you could alter `defineTool` to take a category definition:

```typescript
export type ToolDef<T> = {
  whitelistCategory: WhitelistCategory,
  ArgumentsSchema: t.Type<any>;
  Schema: t.Type<T>;
  validate: (abortSignal: AbortSignal, transport: Transport, t: T, cfg: Config) => Promise<null>;
  run: (
    abortSignal: AbortSignal,
    transport: Transport,
    t: T,
    cfg: Config,
    modelOverride: string | null,
  ) => Promise<ToolResult>;
};

export type ToolFactory<T> = (
  signal: AbortSignal,
  transport: Transport,
  config: Config,
) => Promise<ToolDef<T> | null>;


export function defineTool<T>(factory: ToolFactory<T>): ToolFactory<T> {
  return factory;
}
```

The `defineTool` function only exists to enforce tools defining certain types (as you can see, it doesn't actually do anything — it's just returns whatever you pass in!), so we can use that to also enforce tools have categories. If you move more of this into tool-specific React components (or category-specific React components, that take specific tool types as props), I think you'll find those will take care of the rest of the category stuff :) If someone forgets to add the whitelist handling, TSC will yell at them for not including the UI code somewhere, since there will be a `switch` statement that doesn't return a component somewhere.

source/hooks/use-cwd.tsx
```diff
@@ -0,0 +1,8 @@
+import { createContext, useContext } from "react";
+
+const nullByte = "\0";
+export const CwdContext = createContext<string>(nullByte);
```
sdjidjev, line 4: For the whitelist feature is it worth throwing an error if the output of `useCwd` is the null byte? 

Since the cwd (for the whitelist feature) is only for showing the user that the read/edit permissions only extend to the user's current working directory. But the whitelist key doesn't actually need the cwd at all (since by default it's the entire directory that the user launched `octo` in)

sdjidjev (commented):
@reissbaker Thank you for your reviews!! As we spoke offline, I think removing categories simplifies the feature a bunch.
Updated now to not use categories and applied your feedback!

source/hooks/use-cwd.tsx
reissbaker: Oh interesting! Up to you IMO on this one

PR #82, opened 2026-01-15
Ctrl-p keyboard hotkey extravaganza menu

reissbaker:
This swaps the keyboard shortcut to open the menu to ctrl-p, like OpenCode. Unlike OpenCode, this opens a Starcraft-y ultra keyboard hotkey driven menu, rather than needing to search. I converted nearly every single menu panel in our codebase to be hotkey driven; I skipped the multi-select import model menu for now since it's more of an involved port, and I also skipped the tool request renderer since @sdjidjev is already working on that in a separate PR and I didn't want to cause conflicts. It's a pretty simple port once that change lands though!

I love the feel of the shortcuts. They feel so good.

## Screenshots

A simple hotkey-driven list:
image

Numeric, paginated lists with key hotkeys, using Vim-inspired `h` and `l` shortcuts for navigating forwards and backwards through the list:
image
image

Basic confirmation dialog:
image

### DIFF
--- a/source/components/kb-select/kb-shortcut-select.tsx
+++ b/source/components/kb-select/kb-shortcut-select.tsx
@@ -4,6 +4,7 @@ import { Box, Text, useInput } from "ink";
 import { IndicatorComponent } from "../select.tsx";
 import { useColor } from "../../theme.ts";
 
+// Allowable A-Z hotkeys, minus reserved keys
 export type Hotkey =
   | "a"
   | "b"


source/app.tsx
```diff
@@ -412,7 +419,7 @@ function BottomBarContent({ inputHistory }: { inputHistory: InputHistory }) {
   return (
     <Box flexDirection="column">
       <Box marginLeft={1} justifyContent="flex-end">
-        <Text color="gray">(Press ESC to enter the menu)</Text>
+        <Text color="gray">(Ctrl-p to enter the menu)</Text>
```
sdjidjev: -nit- just for consistency, right now the prompt for mode switching is `Shift+Tab` (`+` vs `-`)
so probably good idea to either change one of them to (`Ctrl+p`) or (`Shift-Tab`)
i have a slight preference for `+` just cuz Claude & VSCode use them

source/components/kb-select/kb-shortcut-select.tsx
```diff
@@ -0,0 +1,223 @@
...
+export type Item<V> = {
+  label: string;
+  value: V;
+};
+type MapShortcutType<V> = {
+  type: "key";
+  mapping: Record<string, Item<V>>;
+};
+type NumberShortcutType<V> = {
+  type: "number";
```
sdjidjev: could we instead make any numbered shortcuts be passed in stringified and as a `MapShortcutType`?

that way we can make the type declaration of ShortcutArray much simpler and also add your pagification code to character-based shortcuts (when there are more than 10 options)

you could write a helper function to help convert situations where you stringify the numbers and output a `ShortcutArray`

```
export function createNumericShortcuts<V>(items: Array<Item<V>>): MapShortcutType<V>[] {
  const pages: MapShortcutType<V>[] = [];
  const totalPages = Math.ceil(items.length / PAGE_SIZE);

  for (let page = 0; page < totalPages; page++) {
    const start = page * PAGE_SIZE;
    const end = Math.min(start + PAGE_SIZE, items.length);
    const pageItems = items.slice(start, end);

    const mapping: Record<string, Item<V>> = {};
    pageItems.forEach((item, index) => {
      mapping[`${index}`] = item;
    });

    pages.push({
      type: "key",
      mapping,
    });
  }

  return pages;
}
```

source/app.tsx
reissbaker: good point! done

source/components/kb-select/kb-shortcut-select.tsx
reissbaker: Oh! So, I should probably write a comment explaining this. My thinking for hotkeys is basically:

* For UI elements controlled by us, we should assign unique a-z hotkeys. Since it's controlled by us, we shouldn't auto-paginate: the UI should just be whatever we design it to be.
* For UI elements that are controlled by user input, where the number of elements can grow/shrink, we can't auto-assign meaningful a-z hotkeys — for example, what if the list grows longer than 26 elements? Also, it starts to look a little crazy at that length and may not fit on a terminal screen. For those elements, we use the index of the item in the user-controlled list as the hotkey, from 0-9. The reason we auto-paginate at 10 entries (as opposed to some other number) is because... after 9, there's no way to distinguish between `1` and `10` unless we add latency and wait for a second keypress! That feels slow and so we paginate there.

The reason for the kind-of-crazy type is that, since we're auto-assigning the hotkeys based on list index, we also can't display multiple user-controlled lists on the same page at the same time, since the hotkeys would conflict: does `0` mean the first item in the first list, or the first item in the second list? While this in theory feels restrictive, in practice I think this actually feels nice: it's a little crazy to imagine multiple numeric-hotkeyed lists open onscreen at the same time and trying to figure out what key to press. But, we *don't* have any restrictions around UI elements that we control: we can have as many or as few of those onscreen as we want. So, basically the tuple type is designed to allow:

* Standalone key mappings
* Standalone user-controlled lists with auto-assigned numeric hotkeys
* A key mapping before the user-controlled list
* A key mapping after the user-controlled list
* Or key mappings surrounding the user-controlled list

Since numeric hotkeys and pagination are only meant for user-controlled lists, I kind of like keeping them separate as types: that way we don't accidentally auto-paginate UI elements that we control, and it also gives us control to make sure there's only ever one of those things onscreen at a time.

reissbaker: Added a comment explaining this and also renamed `numeric` to `auto-list` to better self-document what the intended use cases are!

sdjidjev: that makes sense! thank you for explaining!

PR #89, opened 2026-01-24
Add menu option to clear conversation history

billycao:
Adds an incrementing `clearNonce` to state to trigger re-render of conversation history in `<Static>`.

https://github.com/user-attachments/assets/d46e140c-2696-4d49-acf8-4636636f7112

### DIFF
--- a/source/menu.tsx
+++ b/source/menu.tsx
@@ -23,8 +23,7 @@ type MenuMode =
   | "set-default-model"
   | "quit-confirm"
   | "remove-model"
-  | "clear-confirm"
-  | "context-menu";
+  | "clear-confirm";
 
 type MenuState = {
   menuMode: MenuMode;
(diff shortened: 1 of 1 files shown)


source/menu.tsx
```diff
@@ -526,6 +536,57 @@ function QuitConfirm() {
...
+      toggleMenu: state.toggleMenu,
+      notify: state.notify,
+    })),
+  );
+
+  useInput((_, key) => {
+    if (key.escape) setMenuMode("main-menu");
+  });
+
+  const items: Keymap<"no" | "yes"> = {
```
reissbaker: I think you can replace most of this with the `ConfirmDialog` in `source/components/confirm-dialog.tsx`

billycao: Ah my intern based it on QuitConfirm below so went ahead and updated that too (preserving the no first order)

PR #94, opened 2026-01-28
Session Persistence & Continuation

LeoMassie:
## Overview
This PR adds local session persistence to Octofriend. Before this, you lost all chat history when you exited. Now conversations are stored in a local SQLite database at `~/.config/octofriend/db.sqlite`, so you can browse old sessions and pick up where you left off.

I also cleaned up the TUI while working on this—specifically the screen clearing behavior and the menu system.

## Architecture
We picked SQLite via `better-sqlite3` and `drizzle-orm` for persistence. Reasons:
1. Zero-conf: it's just a file, no server required.
2. Fast for local queries.
3. Drizzle gives us full TypeScript types for the schema.

### Schema
Two new tables:
- `sessions`: stores metadata (name, model, last active timestamp)
- `messages`: stores the chat history (prompts, responses, tool calls)
  - `messages` has a foreign key to `sessions` with `ON DELETE CASCADE`, so deleting a session wipes its messages automatically.

### Persistence
- Auto-save: sessions save as you chat
- Lazy creation: no db entry until you send your first message (prevents empty "ghost" sessions)
- BigInt handling: history objects use BigInt, which JSON doesn't support natively. We added a serialization layer to handle this.

### Session Management
- Resume: list sessions sorted by recency and jump back in
- Wipe: "Wipe All History" option (with confirmation)
- Menu reorg: moved "New Conversation" into a "Sessions" submenu to clean up the main menu

### UI Refinements
- Smart clearing: the terminal clears scrollback when you switch sessions. Makes it feel like opening a new file instead of appending to infinite scroll.
- Banner stability: fixed the ASCII art logo disappearing or duplicating. It now anchors the start of every session consistently.

## Build & Distribution
- CLI permissions: added `chmod +x` to the build script in `package.json`. This ensures `cli.js` has execute permissions when installed globally via `npm install -g .`, preventing "Permission denied" errors.

## Testing
- Manual verification: creating, resuming, deleting sessions works
- Edge cases: tested Ctrl+C exits, empty sessions, and schema migrations on fresh installs
- Migrations: auto-migration runs on boot to keep the SQLite schema current

## Changed Components
- `source/db/`: database setup and error-handled migrations
- `source/state.ts`: Zustand store writes to DB on every message; added session loading/listing
- `source/menu.tsx`: session navigation UI, reorganized main menu
- `source/sessions/`: schema for sessions and messages
- `source/app.tsx` & `source/cli.tsx`: session loading and boot-time migrations
- `package.json`: build script update for CLI stability

### DIFF
--- a/source/cli.tsx
+++ b/source/cli.tsx
@@ -152,8 +152,10 @@ async function runMain(opts: { config?: string; unchained?: boolean; transport: 
         updates={await readUpdates()}
         inputHistory={await loadInputHistory()}
       />,
+      {
+        exitOnCtrlC: false,
+      },
     );
-
     await waitUntilExit();
 
     console.log("\nApprox. tokens used:");


source/db/migrate.ts
```diff
@@ -10,7 +10,13 @@ export async function migrate() {
       ? path.join(__dir, "../../drizzle/")
       : path.join(__dir, "../../../drizzle/");
 
-  drizzleMigrate(db(), {
-    migrationsFolder: migrationsPath,
-  });
+  try {
```
reissbaker, line 13: I think this try/catch is probably unnecessary since it's re-throwing the error?

source/sessions/schema/session-schema.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const messagesTable = sqliteTable("messages", {
+  id: integer().primaryKey({ autoIncrement: true }),
+  sessionId: integer()
+    .notNull()
+    .references(() => sessionsTable.id, { onDelete: "cascade" }),
+  sequenceId: integer().notNull(), // To match HistoryItem.id
+  data: text().notNull(), // JSON blob of HistoryItem
+  createdAt: integer({ mode: "timestamp" }).notNull(),
+});
```
reissbaker, line 18: I would recommend adding an index here:

```typescript
export const messagesTable = sqliteTable("messages", {
  // ...
}, table => [
  index().on(table.sessionId),
]);
```

source/sessions/schema/session-schema.ts
```diff
@@ -0,0 +1,18 @@
...
+  model: text().notNull(),
+});
+
+export const messagesTable = sqliteTable("messages", {
+  id: integer().primaryKey({ autoIncrement: true }),
+  sessionId: integer()
+    .notNull()
+    .references(() => sessionsTable.id, { onDelete: "cascade" }),
+  sequenceId: integer().notNull(), // To match HistoryItem.id
+  data: text().notNull(), // JSON blob of HistoryItem
```
reissbaker, line 16: Definitely prefer to have a typed schema here rather than JSON blobs, so that we can manage changes via migrations. With JSON blobs, making changes to LlmIR/HistoryItems becomes pretty tough: we have to carefully manage the history of all previous types in the codebase.

One way to do this would be to have one table per HistoryItem type. Since we have a `sequenceId` to order them, we could construct the history by constructing N queries (one per table), and then sorting them in-memory by sequence ID. Since the sequence ID is sequential, in the future we could also do nice optimizations like fetching from the compaction checkpoint table first, and only fetching messages with `sequenceId > MOST_RECENT_CHECKPOINT_SEQUENCE_ID` (although this seems like a future optimization option and not something we need right now!).

source/cli.tsx
```diff
@@ -32,6 +32,7 @@ import { migrate } from "./db/migrate.ts";
 import { run } from "./compilers/run.ts";
 import { loadInputHistory } from "./input-history/index.ts";
 import { makeAutofixJson } from "./compilers/autofix.ts";
+import { useAppStore } from "./state.ts";
```
reissbaker, line 35: Seems unused?

source/menu.tsx
```diff
@@ -737,3 +751,185 @@ function AddModelMenuFlow() {
...
+  // C (index 2) => Hotkey 1.
+  const totalSessions = sessions.length;
+  const numericItems = sessions.map((session, index) => {
+    const chronoNumber = totalSessions - index;
+    // Use modulo 10 for display key if > 9.
+    const key = String(chronoNumber <= 9 ? chronoNumber : chronoNumber % 10);
+    return {
+      label: `${session.name} - ${session.lastActiveAt.toLocaleString()}`,
+      value: `session-${session.id}` as const,
+      hotkey: key,
```
reissbaker, line 843: You don't need to add this `hotkey` field: the `auto-list` type already creates numeric, paginated hotkeys for everything passed in.

Instead you can sort the sessions using `sessions.sort(...)` and then transform the sorted array into `Item<V>` via `.map(...)`.

source/state.ts
```diff
@@ -468,8 +524,113 @@ export const useAppStore = create<UiState>((set, get) => ({
...
+  wipeAllHistory: async () => {
+    const { history } = get();
+    await db().delete(schema.sessionsTable);
+    // Messages will be deleted by cascade
+
+    // Only update state if there's actually history to clear
+    if (history.length > 0) {
+      // Clear screen before state update
+      if (process.stdout.isTTY) {
+        process.stdout.write("\x1b[2J\x1b[3J\x1b[H");
```
reissbaker, line 597: This will cause some unexpected behavior if e.g. Octo didn't previously take up the entire terminal window — instead, use the `clearNonce` to clear the rendered history. In fact, I think you already are using `clearNonce`: you can probably just delete this line!

source/menu.tsx
```diff
@@ -737,3 +751,185 @@ function AddModelMenuFlow() {
...
+    useShallow(state => ({
+      wipeAllHistory: state.wipeAllHistory,
+      toggleMenu: state.toggleMenu,
+      notify: state.notify,
+    })),
+  );
+
+  return (
+    <ConfirmDialog
+      confirmLabel="Yes, WIPE ALL HISTORY"
```
reissbaker, line 895: Haha, I think this is a little scary. Rather than having a menu item to wipe all history, I think we'd want to be able to delete specific sessions.

source/state.ts
```diff
@@ -408,8 +444,28 @@ export const useAppStore = create<UiState>((set, get) => ({
...
       set({ history: [...historyCopy] });
+
+      const sessionId = await get()._ensureSession(config);
+      if (sessionId != null) {
+        for (const msg of outputMessages) {
+          await db()
+            .insert(schema.messagesTable)
+            .values({
+              sessionId,
+              sequenceId: Number(msg.id),
```
reissbaker, line 458: Hmmm. Since SQLite doesn't have a native bigint, and Drizzle annoyingly uses Blobs for BigInts in SQLite (which means you can't compare them via sql queries), we should probably just make `sequenceId` a `number` rather than a `bigint` and store it as a `REAL` in SQLite.

source/state.ts
```diff
@@ -468,8 +524,113 @@ export const useAppStore = create<UiState>((set, get) => ({
...
+    // Abort any ongoing responses to avoid polluting the new cleared state.
+    const { abortResponse, history } = get();
+    abortResponse();
+
+    // Only update state if there's actually history to clear
+    // This prevents unnecessary Static component remounts that cause duplicate intros
+    if (history.length > 0) {
+      // Clear screen before state update
+      if (process.stdout.isTTY) {
+        process.stdout.write("\x1b[2J\x1b[3J\x1b[H");
```
reissbaker, line 538: This also seems unnecessary — that's what `clearNonce` is for!

source/state.ts
```diff
@@ -468,8 +524,113 @@ export const useAppStore = create<UiState>((set, get) => ({
...
+      }
+      set({
+        currentSessionId: null,
+        history: [],
+        clearNonce: get().clearNonce + 1,
+      });
+    }
+  },
+
+  _ensureSession: async (config: Config) => {
```
reissbaker, line 607: Rather than calling `ensureSession` all the time, we should probably create a session on boot and pass it directly into the App component (and from there, the Zustand state).

source/state.ts
```diff
@@ -468,8 +524,113 @@ export const useAppStore = create<UiState>((set, get) => ({
...
+    }
+
+    set({
+      currentSessionId: sessionId,
+      history,
+      clearNonce: get().clearNonce + 1,
+    });
+  },
+
+  listSessions: async () => {
```
reissbaker, line 574: My approach for async data in React/Zustand is: load the data upfront, and then just manage it in-memory. Otherwise we have to deal with empty states / flashes of content / loading states. In this case, I think it's worth loading the recent sessions in `cli.tsx` and then passing them into the `App` component.

source/state.ts
```diff
@@ -468,8 +524,113 @@ export const useAppStore = create<UiState>((set, get) => ({
...
+      .select()
+      .from(schema.messagesTable)
+      .where(eq(schema.messagesTable.sessionId, sessionId))
+      .orderBy(schema.messagesTable.id);
+
+    const history: HistoryItem[] = messages.map(m => deserializeHistoryItem(m.data));
+
+    // Clear screen before state update
+    if (process.stdout.isTTY) {
+      process.stdout.write("\x1b[2J\x1b[3J\x1b[H");
```
reissbaker, line 564: Ditto re: `clearNonce` should handle this

reissbaker (changes requested):
Thanks for working on such a useful feature! It's been on both my and @sdjidjev's plates to build something like this for a while now :)

Left some PR feedback — I think this is a reasonable start on the problem. If you're able to keep working on it definitely feel free to, or we can also use this PR as a starting point if you're too busy!

Issue #43, opened 2025-10-17
Input UX improvement request


casparvitch:
It would be nice to be able to Ctrl+arrow navigate through the input text, home/end buttons, ...

PR #96, opened 2026-01-29
Add Home/End and Ctrl+arrow text input navigation

bradfa:
The Home/End key input in ink depends on ink >= v6.6.0 as that was the first release to support it.  Octofriend moved to ink v6.6.0 in fbc7ddffbefd1da51c00df95b8e017a3dabfc44a.

Fixes: https://github.com/synthetic-lab/octofriend/issues/43


reissbaker (approved):
Amazing!

PR #97, opened 2026-01-30
Prevent key strokes after control to leak into input.

sdjidjev:
When I did Ctrl+P earlier to open the menu, I noticed that it added a "p" in the input, so decided to blanket ignore any characters typed after "Ctrl".

The question is, I don't use Vim; would this fix mess things up? I didn't try it yet but...

### DIFF
--- a/source/components/text-input.tsx
+++ b/source/components/text-input.tsx
@@ -191,13 +191,7 @@ export default function TextInput({
         // Vim didn't consume it, continue with normal processing
       }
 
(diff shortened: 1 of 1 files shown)


reissbaker:
@bradfa did a targeted fix that should also work with Vim mode here: https://github.com/synthetic-lab/octofriend/pull/107 :grin: Do you feel like it's still worth merging this, or should we close?

Issue #95, opened 2026-01-29
Multi-line vim 0 and $ do not comprehend the multi-line situation


bradfa:
In vim mode, when inputting multi-line text by the user pressing the Enter key to insert a newline, then when navigating in command mode pressing 0 will take you to the very first character in the text box and $ will take you to the very last.  This movement may cross line boundaries.

This is not how actual multi-line editing in vim works, where you would be taken to the first or last character **on the current line**.

For example, if this was your text input box:

```
This is octofriend (you! but you already knew this) and we need to fix
what happens when the user is in vim mode and presses the "0" and "$" keys
so that they comprehend multi-line text box movement.

"0" should move to the first char in the CURRENT line.  "$" should move to
the last in the CURRENT line.

The existing implementation moves to the first or last character in the entire
text box, which is not how vim works.
```

Then currently you press 0 anywhere in that text input and you'll move to the "T" which starts "This".  Press "$" and you'll move to the last period.  This is not correct.

PR #100, opened 2026-01-30
vim-mode command and motion fixes for handling newlines

bradfa:
- Fixes `0`, `$`, `^`, `I`, `a`, and improves `d` commands for handling newlines.
- Fixes `$` motion for handling newlines.
- Adds `D` command (equivalent to `d$`).

Fixes: https://github.com/synthetic-lab/octofriend/issues/95

### DIFF
--- a/source/components/vim-mode.tsx
+++ b/source/components/vim-mode.tsx
@@ -336,7 +336,14 @@ export function useVimKeyHandler(
         if (key.escape || (key.ctrl && input === "c")) {
           let newCursorPosition = cursorPosition;
           if (cursorPosition > 0) {
-            newCursorPosition = cursorPosition - 1;
+            // Special case: if we're at the start of an empty line, stay there
+            // (empty line = only valid position is column 0)
+            const currentLineInfo = getLineInfo(currentValue, cursorPosition);
+            const lineStart = getLineStart(currentValue, currentLineInfo.lineIndex);
+            const currentLine = getLineText(currentValue, currentLineInfo.lineIndex);
+            if (!(cursorPosition === lineStart && currentLine.length === 0)) {
+              newCursorPosition = cursorPosition - 1;
+            }
           }
           if (insertStartStateRef.current !== null) {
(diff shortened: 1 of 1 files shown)


source/components/vim-mode.tsx
```diff
@@ -165,32 +211,43 @@ const motions: Record<string, Motion> = {
     return { start: cursorPosition, end: 0 };
```
sdjidjev: i think you'll also have to update `end` here as well to calculate `lineStart`

source/components/vim-mode.tsx
```diff
@@ -55,11 +64,48 @@ const getLineStart = (text: string, lineIndex: number): number => {
   return position;
 };
 
+const getLineEnd = (text: string, lineIndex: number): number => {
+  const lineStart = getLineStart(text, lineIndex);
+  const lines = text.split("\n");
+  const lineLength = lines[lineIndex]?.length || 0;
+  return lineStart + lineLength - 1;
```
sdjidjev: For empty lines we should use: `lineStart + Math.max(0, lineLength - 1)`

source/components/vim-mode.tsx
sdjidjev: we'll need to update this to check against `lineStart` instead of `0`.

sdjidjev (changes requested):
@bradfa This will be a huge QoL improvement for our vim users, thank you! 
Just a few more edge cases need handling and then should be good to go!

source/components/vim-mode.tsx
bradfa: Ah, yes, as otherwise 'h' will "move left" crossing newlines.  Fix incoming.

source/components/vim-mode.tsx
bradfa: Yep, the motion should match the command for '0'.

source/components/vim-mode.tsx
bradfa: Agreed!

Issue #55, opened 2025-11-20
When using a Synthetic-provided model, show subscription quota usage


bradfa:
Similar to how this VScode extension (https://marketplace.visualstudio.com/items?itemName=nrw.vscode-synthetic-quota) works where it shows your Synthetic subscription quota usage, if Octo knows we're using a Synthetic-provided model and can query to find out if the current API key is on a subscription plan, then it would be super handy to show the current subscription quota usage directly in Octofriend.

If this could live-update after each user->agent back and forth and be shown below the text input box to show the usage and optionally also show how much time is left until the quota resets, that would be quite slick.

Something like:

```
Quota: 1.1/135 req, Reset: 4h
```

PR #105, opened 2026-01-30
Implement Synthetic quota indication

bradfa:
Show the current quota use for requests, searches, and tool calls below the text input area.  Refreshes every time control is given back to the user.  Is hidden when the software update result is shown.

Due to how tool calls get credited after the fact, if the agent does a bunch of tool calls then the requests quota usage shown my reflect the instantaneous value which may be higher than reality after the tool call crediting happens, and conversely, the tool call quota may be slightly lower than reality.  But the goal is to give insight so users are aware of roughly where they're at with their subscription so this is a reasonable trade-off.

Fixes: https://github.com/synthetic-lab/octofriend/issues/55

### DIFF
--- a/source/components/menu-quota-indicator.tsx
+++ b/source/components/menu-quota-indicator.tsx
@@ -3,6 +3,7 @@ import { Text, Box } from "ink";
 import { assertKeyForModel, useConfig } from "../config.ts";
 import { useModel } from "../state.ts";
 import { t } from "structural";
+import { formatTimeUntil } from "../time.ts";
 
 const QuotaResponseSchema = t.subtype({
   subscription: t.subtype({
@@ -19,34 +20,6 @@ type QuotaData = {
   limit: number;
   renewsAt: Date;
 };
-
-function formatTimeUntil(renewsAt: Date): string {
-  const now = new Date();
-  const diffMs = renewsAt.getTime() - now.getTime();
-  const diffMins = Math.floor(diffMs / (1000 * 60));
-
-  if (diffMins < 60) {
-    return `in ${diffMins} minute${diffMins !== 1 ? "s" : ""}`;
-  }
-
-  const diffHours = Math.floor(diffMins / 60);
-  const remainingMins = diffMins % 60;
-
-  if (diffHours < 24) {
-    if (remainingMins > 0) {
-      return `in ${diffHours} hour${diffHours !== 1 ? "s" : ""} ${remainingMins} minute${remainingMins !== 1 ? "s" : ""}`;
-    }
-    return `in ${diffHours} hour${diffHours !== 1 ? "s" : ""}`;
-  }
-
-  const diffDays = Math.floor(diffHours / 24);
-  const remainingHours = diffHours % 24;
-
-  if (remainingHours > 0) {
-    return `in ${diffDays} day${diffDays !== 1 ? "s" : ""} ${remainingHours} hour${remainingHours !== 1 ? "s" : ""}`;
-  }
-  return `in ${diffDays} day${diffDays !== 1 ? "s" : ""}`;
-}
 
 async function fetchQuota(apiKey: string): Promise<QuotaData | null> {
   try {
@@ -108,7 +81,7 @@ export const MenuQuotaIndicator = React.memo(() => {
   if (loading) {
     return (
       <Box flexDirection="column" alignItems="center">
-        <Text color="gray">Synthetic Subscription:</Text>
+        <Text color="gray">Synthetic Subscription</Text>
         <Text color="gray">Loading quota...</Text>
       </Box>
     );
--- a/source/time.test.ts
+++ b/source/time.test.ts
@@ -1,0 +1,111 @@
+import { describe, it, expect } from "vitest";
+import { formatTimeUntil } from "./time.ts";
+
+describe("formatTimeUntil", () => {
+  const now = new Date("2026-03-02T12:00:00Z");
+
+  it("should format minutes only (singular)", () => {
+    const expiresAt = new Date("2026-03-02T12:01:00Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 1 minute");
+  });
+
+  it("should format minutes only (plural)", () => {
+    const expiresAt = new Date("2026-03-02T12:30:00Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 30 minutes");
+  });
+
+  it("should format hours only (singular)", () => {
+    const expiresAt = new Date("2026-03-02T13:00:00Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 1 hour");
+  });
+
+  it("should format hours only (plural)", () => {
+    const expiresAt = new Date("2026-03-02T15:00:00Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 3 hours");
+  });
+
+  it("should format hours and minutes", () => {
+    const expiresAt = new Date("2026-03-02T13:45:00Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 1 hour 45 minutes");
+  });
+
+  it("should format days only (singular)", () => {
+    const expiresAt = new Date("2026-03-03T12:00:00Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 1 day");
+  });
+
+  it("should format days only (plural)", () => {
+    const expiresAt = new Date("2026-03-05T12:00:00Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 3 days");
+  });
+
+  it("should format days and hours", () => {
+    const expiresAt = new Date("2026-03-03T18:00:00Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 1 day 6 hours");
+  });
+
+  it("should handle 59 minutes as minutes", () => {
+    const expiresAt = new Date("2026-03-02T12:59:00Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 59 minutes");
+  });
+
+  it("should handle exactly 60 minutes as 1 hour", () => {
+    const expiresAt = new Date("2026-03-02T13:00:00Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 1 hour");
+  });
+
+  it("should handle 23 hours as hours", () => {
+    const expiresAt = new Date("2026-03-03T11:00:00Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 23 hours");
+  });
+
+  it("should handle exactly 24 hours as 1 day", () => {
+    const expiresAt = new Date("2026-03-03T12:00:00Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 1 day");
+  });
+
+  it("should handle 1 hour 1 minute correctly", () => {
+    const expiresAt = new Date("2026-03-02T13:01:00Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 1 hour 1 minute");
+  });
+
+  it("should handle 1 day 1 hour correctly", () => {
+    const expiresAt = new Date("2026-03-03T13:00:00Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 1 day 1 hour");
+  });
+
+  it("should round up less than 1 minute to 1 minute", () => {
+    const expiresAt = new Date("2026-03-02T12:00:30Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 1 minute");
+  });
+
+  it("should round up 1 second to 1 minute", () => {
+    const expiresAt = new Date("2026-03-02T12:00:01Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 1 minute");
+  });
+
+  it("should handle 0 time difference as 0 minutes", () => {
+    const expiresAt = new Date("2026-03-02T12:00:00Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 0 minutes");
+  });
+
+  it("should handle negative seconds as 0 minutes", () => {
+    const expiresAt = new Date("2026-03-02T11:59:30Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 0 minutes");
+  });
+
+  it("should handle negative minutes as 0 minutes", () => {
+    const expiresAt = new Date("2026-03-02T11:55:00Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 0 minutes");
+  });
+
+  it("should handle negative hours as 0 minutes", () => {
+    const expiresAt = new Date("2026-03-02T10:00:00Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 0 minutes");
+  });
+
+  it("should handle negative days as 0 minutes", () => {
+    const expiresAt = new Date("2026-02-28T12:00:00Z");
+    expect(formatTimeUntil(expiresAt, now)).toBe("in 0 minutes");
+  });
+});
--- a/source/time.ts
+++ b/source/time.ts
@@ -1,0 +1,31 @@
+export function formatTimeUntil(expiresAt: Date, now: Date = new Date()): string {
+  const diffMs = expiresAt.getTime() - now.getTime();
+
+  if (diffMs <= 0) {
+    return "in 0 minutes";
+  }
+
+  const diffMins = Math.max(1, Math.ceil(diffMs / (1000 * 60)));
+
+  if (diffMins < 60) {
+    return `in ${diffMins} minute${diffMins !== 1 ? "s" : ""}`;
+  }
+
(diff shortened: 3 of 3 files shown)


source/components/synthetic-quota-indicator.tsx
```diff
@@ -0,0 +1,203 @@
+import React, { useEffect, useState, useRef } from "react";
+import { Text } from "ink";
+import { useConfig, assertKeyForModel } from "../config.ts";
+import { useModel } from "../state.ts";
+import { providerForBaseUrl } from "../providers.ts";
+
+type QuotaResponse = {
+  subscription: {
+    limit: number;
```
sdjidjev: you should make a `QuotaResponseEntry` to reduce repetition here:

```
  type QuotaResponseEntry = {                                                                                                      
    limit: number;                                                                                                                 
    requests: number;                                                                                                              
    renewsAt: string;                                                                                                              
  };                                                                                                                               
                                                                                                                                   
  type QuotaResponse = {                                                                                                           
    subscription: QuotaResponseEntry;                                                                                              
    search: {                                                                                                                      
      hourly: QuotaResponseEntry;                                                                                                  
    };                                                                                                                             
    toolCalls: QuotaResponseEntry;                                                                                                 
  };
```

source/components/synthetic-quota-indicator.tsx
```diff
@@ -0,0 +1,203 @@
...
+  error: boolean;
+};
+
+export type { QuotaData };
+
+export function useSyntheticQuotaData(refreshTrigger: number): QuotaData | null {
+  const config = useConfig();
+  const model = useModel();
+  const [quota, setQuota] = useState<QuotaData>({
+    subscription: { used: 0, limit: 0, renewsAt: null },
```
sdjidjev: you should make a constant `DEFAULT_QUOTA_ENTRY` for these values

source/components/synthetic-quota-indicator.tsx
```diff
@@ -0,0 +1,203 @@
...
+    search: { used: 0, limit: 0, renewsAt: null },
+    toolCalls: { used: 0, limit: 0, renewsAt: null },
+    loading: true,
+    error: false,
+  });
+  const isFetchingRef = useRef(false);
+
+  const provider = providerForBaseUrl(model.baseUrl);
+  const isSynthetic = provider?.name === "Synthetic";
```
sdjidjev: we already have a `SYNTHETIC_PROVIDER` constant that you can get the `name` attribute from.

source/app.tsx
```diff
@@ -260,7 +269,11 @@ function BottomBar({
 
   return (
     <Box flexDirection="column" width="100%">
-      <BottomBarContent inputHistory={inputHistory} />
+      <BottomBarContent
```
sdjidjev: We were discussing this feature earlier and think a better placement for it might be underneath the `Collaboration` vs `Unchained` mode indicator
Image
Image
It appears on boot & when modes switch, and then disappears after 5 seconds. There's a whole line worth of space there so even on smaller terminals we can see the full status.

source/components/synthetic-quota-indicator.tsx
```diff
@@ -0,0 +1,203 @@
...
+      search: { used: 0, limit: 0, renewsAt: null },
+      toolCalls: { used: 0, limit: 0, renewsAt: null },
+      loading: false,
+      error: true,
+    };
+  }
+}
+
+function formatRenewsAt(date: Date): string {
+  const hours = date.getHours();
```
sdjidjev: Maybe worth just adding a dependency to a date formatter. There's a lightweight one like [dayjs](https://www.npmjs.com/package/dayjs). I'd be in favor of just adding that, but wanna check what @reissbaker thinks.

sdjidjev (changes requested):
This is a feature we've been wanting to add to Octo for a while! Thanks for putting in the work to get the ball rolling!

Did a quick skim and added a few requested changes.

Wanted to do a more thorough review, so I pulled the branch, but I can't get the quota component to show up for me.

I saw you posted a screenshot on Discord, would also love more screenshots on your PR for posterity 🙌  With my initial skim it's not immediately obvious to me how it behaves with showing all the different quota modes, so it would help to have a reference with your screenshots and the intended behaviors.

Thanks again for your contributions and support to Synthetic and Octo! 🐙

source/components/synthetic-quota-indicator.tsx
bradfa: Yeah, probably a good idea.  I didn't want to bring in a new dependency and when I did test once for real hitting the quota limit for the subscription requests, I got the billing error and Octo wouldn't update the requests (I presume because the transition from a not-input mode to input mode wasn't happening).  But restarting Octo would show a time formatted like "12:45pm" for when the quota would reset.

Maybe it would be better to just show how much time until the reset happens?  Then rendering datetime and time zone things goes away and we just calculate the difference between the reset time and the current time, with all times in UTC?

source/app.tsx
bradfa: Yep.  If you try again with my extra commit I added to this PR, to fix the API rename, it should work for you now.  When the version check hides after the 5 seconds then it'll show the quota data if you are currently using a Synthetic model.  If you're not using a Synthetic model then it won't show the quota info.

bradfa: I'm not opposed to moving the quota info below the collaboration/unchained mode indicator, but I didn't want to add another line of status info if I didn't have to, as there's already an extra 5-second transient line which explains the collaboration/unchained mode meaning every time you switch modes.  Putting it on the right and only showing when the version check string isn't showing seemed like a reasonable tradeoff to me in terms of not adding an extra line and being visible most of the time.

But if there's a consensus to move it somewhere else, that's fine, too.  Having it hide when the version check displays and having it move up/down when entering/exiting vim insert mode was non-trivial to figure out, although it works pretty reliably now.

sdjidjev: Thanks for being considerate of using space efficiently!
Yep, we already discussed amongst the team a while back that we wanted to add it at that empty space below collaboration/unchained mode, since it remains empty as long as the 5-second transient line isn't there.

source/components/synthetic-quota-indicator.tsx
```diff
@@ -0,0 +1,204 @@
...
+    </Text>
+  );
+}
+
+export const SyntheticQuotaIndicator = React.memo(({ quota }: { quota: QuotaData | null }) => {
+  if (!quota || quota.loading || quota.error) return null;
+
+  return (
+    <Text color="gray">
+      <QuotaLabel label="R" quota={quota.subscription} />{" "}
```
sdjidjev: Hmm now looking at the screenshots, I feel like the `R`/`S`/`T` labels are not immediately obvious to what kind of value they're indicating. Tonight I'm going to think about the best way to show all the data, or just choose one quota data to show (or maybe some option to toggle data types, idk) 🤔 I'll let you know sometime this weekend.
The good news is that in the new proposed placement, we will have a lot more room to display text!

bradfa: I debated the right way to do it. None seemed ideal. Totally open to suggestions!

source/components/synthetic-quota-indicator.tsx
reissbaker: Oh, I like the idea of showing how much time is left until the reset happens! I think it's a little easier to read quickly (I'm not always hyper aware of what time it is while I'm coding, so "In 15 minutes" would help me vs an explicit time), and it means we don't have to think about things like "what is the user's current timezone setting and do they use 24hr vs 12hr clocks" etc.

bradfa:
> FYI — we now also send `x-synthetic-quotas` headers down on every OpenAI/Anthropic API request, which include the same data as the `/quotas` endpoint as a JSON string. Might be nice to parse those out of the response in the `standard.ts` compiler (if the headers exist)

I didn't realize this existed, but yes, that's definitely a good idea since Octo already receives that data.

I think it might still be good to poll the API at some regular cadence, as I often have very long-lived (multi-day, sometimes over a week) sessions for specific tasks and it would be nice to have a reasonably up to date indication in every Octo instance I'm running.  Maybe poll at like 5 minute intervals if the quota data hasn't been updated in the past 5 minutes?

And if these 5-minute polls fail, just ignore the failure and keep showing the old data.  Like if the user goes out of wifi range so their network drops, no need to show errors in the quota info.

source/components/synthetic-quota-indicator.tsx
bradfa: New style is like:

```
Req:4/135 Srch:0/250 Tool:79/1620
```

Do you think that's descriptive enough?  I'm trying to balance keeping it short but still being readable.

source/components/synthetic-quota-indicator.tsx
bradfa: Fixed.

source/app.tsx
bradfa: Moved to below the collaboration/unchained mode indicator line.

source/components/synthetic-quota-indicator.tsx
bradfa: Fixed but untested as I don't have a mock API setup currently to send back an API response where I've exceeded the quota and I haven't hit my quota since implementing the change.  It should show like `In 15m` to indicate that the quota will reset in 15 minutes.

source/components/synthetic-quota-indicator.tsx
```diff
@@ -0,0 +1,261 @@
...
+      () => {
+        const age = Date.now() - lastUpdated;
+        const STALE_THRESHOLD = 5 * 60 * 1000; // 5 minutes
+
+        if (age > STALE_THRESHOLD) {
+          loadQuota();
+        }
+      },
+      5 * 60 * 1000,
+    ); // Check every 5 minutes
```
sdjidjev: hmm, i think it's unlikely the menu will remain open for longer than 5 minutes.
let's simplify the component by removing the background refresh check. i think the initial poll is plenty,

source/components/synthetic-quota-indicator.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+  subscribe(listener: (data: QuotaData) => void): () => void {
+    this.listeners.add(listener);
+    return () => {
+      this.listeners.delete(listener);
+    };
+  },
+};
+
+export function updateQuotaFromHeader(headerValue: string): void {
```
sdjidjev: without the need for background refresh we can remove this function as well as the `quotaManager`

source/compilers/standard.ts
```diff
@@ -17,6 +17,7 @@ import { compactionCompilerExplanation } from "./autocompact.ts";
 import { ToolDef } from "../tools/common.ts";
 import { JsonFixResponse } from "../prompts/autofix-prompts.ts";
 import * as irPrompts from "../prompts/ir-prompts.ts";
+import { updateQuotaFromHeader } from "../components/synthetic-quota-indicator.ts";
```
sdjidjev: since this file is in the compiler layer, it shouldn't import from a UI component.

instead we should probably extract `updateQuotaFromHeader`, related functions, and schemas to a separate `source/quota-data.ts` file and have `standard.ts` and `synthetic-quota-indicator.tsx` import from there.

source/components/synthetic-quota-indicator.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+export function useQuotaData(): QuotaData | null {
+  const context = useContext(QuotaContext);
+  if (!context) {
+    return null;
+  }
+  return context.quotaData;
+}
+
+export function useSetQuotaData(): (data: QuotaData) => void {
```
sdjidjev: unused code?

source/components/synthetic-quota-indicator.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+const DEFAULT_QUOTA_ENTRY: QuotaEntry = { used: 0, limit: 0, renewsAt: null };
+
+type QuotaData = {
+  subscription: QuotaEntry;
+  loading: boolean;
+  error: boolean;
+};
+
+export type { QuotaData };
```
sdjidjev: - nit - just export `QuotaData` where you're declaring it!
```
export type QuotaData = {
  subscription: QuotaEntry;
  loading: boolean;
  error: boolean;
};
```

source/components/synthetic-quota-indicator.tsx
```diff
@@ -0,0 +1,261 @@
...
+    return `in ${diffDays} day${diffDays !== 1 ? "s" : ""} ${remainingHours} hour${remainingHours !== 1 ? "s" : ""}`;
+  }
+  return `in ${diffDays} day${diffDays !== 1 ? "s" : ""}`;
+}
+
+/**
+ * Simplified quota display for the main menu.
+ * Shows requests used/limit and refresh time.
+ */
+export const MenuQuotaIndicator = React.memo(({ quota }: { quota: QuotaData | null }) => {
```
sdjidjev: rename `synthetic-quota-indicator.tsx` to match the main component within it (ie `menu-quota-indicator.tsx`)

source/components/synthetic-quota-indicator.tsx
```diff
@@ -0,0 +1,261 @@
...
+  } catch {
+    return {
+      subscription: DEFAULT_QUOTA_ENTRY,
+      loading: false,
+      error: true,
+    };
+  }
+}
+
+function formatTimeUntil(renewsAt: Date): string {
```
sdjidjev: nice util function! typically i'd prefer using a library for calculations like this, but most libraries that deal with JS Dates are kind of large, especially if all we need from it is one function.

let's keep this function but move it into a new file `source/time.ts`. use ai to write a couple of tests to make sure it works as expected! if we end up needing more datetime functions maybe we'll switch to a library, but for now i think this is the only use case in octo.

source/components/synthetic-quota-indicator.tsx
```diff
@@ -0,0 +1,261 @@
...
+  const contextValue: QuotaContextValue = {
+    quotaData: isSynthetic ? quotaData : null,
+    lastUpdated: isSynthetic ? lastUpdated : null,
+    setQuotaData: updateQuota,
+  };
+
+  return <QuotaContext.Provider value={contextValue}>{children}</QuotaContext.Provider>;
+}
+
+export function useQuotaData(): QuotaData | null {
```
sdjidjev: lets move this hook and associated QuotaContext into a new `source/hooks/use-quota-data.ts` file

source/app.tsx
```diff
@@ -204,27 +205,29 @@ export default function App({
...
-                          )}
-                        <BottomBar
-                          inputHistory={inputHistory}
-                          metadata={metadata}
-                          tempNotification={tempNotification}
-                        />
-                      </Box>
-                    </TerminalSizeTracker>
-                  </ExitOnDoubleCtrlC>
+                  <QuotaProvider>
```
sdjidjev: let's move this to instead surround the MainMenu component so that it doesn't have to mount until necessary!
In `menu.tsx`:
```
if (menuMode === "main-menu") return <QuotaProvider><MainMenu /></QuotaProvider>;
```

source/components/synthetic-quota-indicator.tsx
bradfa: The intention was more that if a user has multiple tools or multiple Octofriends running, that we want to show fairly recent quota info if the user stops using this Octo and then opens the menu to see the current quota status.

I often leave an agentic tool open for days, sometimes I don't interact with a given session in a tool even over the weekend, it just sits there in a tmux waiting for me to return.  I didn't want to show old stale quota data from the last API request but to have fairly up-to-date quota information.

Possibly in my refactoring of how things work I've mistakenly set this up so it's only checking if the menu has been open for 5 minutes?  I'll have a look.

sdjidjev: ah i see! the way you have it set up definitely still fetches when the menu is closed (every 5 minutes).
but i'm just suggesting that we just remount the quota provider any time the MainMenu opens instead and thus freshly request the latest data as soon as the MainMenu is open.

it'll be the same behavior as https://synthetic.new/billing works right now. the quota doesn't refresh until the user refreshes the page. i think most people will expect that to be the behavior rather than a real-time updated component, and hopefully that works for your use case!

bradfa: OK, gotcha.  I had avoided fetching the updated quota on-demand as it adds latency to opening the menu, which at times had been noticeable to me.  Maybe fetching the quota data could be async and show a placeholder until the data is available so it doesn't delay the menu being drawn?  I hadn't tried that.

But if it can be made async to not delay drawing the menu, I think that would be the best solution.

bradfa: > i think most people will expect that to be the behavior rather than a real-time updated component, and hopefully that works for your use case!

The claude.ai dashboard uses some fancy javascript to update the user's quota information quite regularly, without requiring a page refresh.  That's really my only other experience with such a quota web interface, and it's handy because their settings/billing pages are quite slow to load.  Synthetic's billing page is much faster to load for me.

sdjidjev: The async solution you're suggesting sounds great! And yeah a placeholder while fetching the quota would be appreciated as to not move the UI around too much once it's done loading. Excited to see it!

sdjidjev (approved):
Fantastic! Looks awesome, thank you for implementing this highly requested feature!

PR #109, opened 2026-01-31
feat: add command-based API key authentication

Amolith:
Give users the option to provide Octo a command, like pass, op (1Password), etc. to obtain their API key instead of writing it to plain text files or long-lived environment variables accessible to more programs than necessary.

- Add new Auth type with 'env' and 'command' variants
- Add shell-quote dependency for safe command parsing
- Update model flows to support command-based auth
- Add retry UI for failed auth commands
- Maintain backward compatibility with existing env var auth

Marked as a draft for now because I want to test it some more first. Feedback is welcome now, after I mark it ready, whenever!


reissbaker (approved):
Looks great!

PR #113, opened 2026-02-01
feat: add @ file autocomplete suggestions

### LINKED ISSUE #110: Feature Request: File Path Auto-complete with Fuzzy Search in Terminal
Add intelligent file path auto-completion with fuzzy search to Octo. When users type `@` followed by text, a dropdown with relevant file/folder suggestions appears.

User Workflow:
1. User types `@` to trigger suggestions
2. A dropdown appears with files/directories from the current or specified directory
3. Arrow keys navigate, Tab/Enter to select
4. Selected path gets inserted at cursor position

This would be a super useful feature that complements Octo's file-system-awareness and makes it more efficient to reference code files during your coding sessions! Let me know what you think. 

I would love to make a PR for this : )

Image

tanvesh01:
## Summary
Add `@file` autocomplete in the input, with gitignore-aware file search and keyboard-driven selection.
 ### UX
- Type `@` to open file suggestions
- Arrow keys (↑/↓) navigate; Enter selects; Esc dismisses
- Suggestions stay open only while the token after `@` looks like a path (no spaces / invalid chars)
- Selecting a suggestion inserts `@<path>` in the editor and appends a trailing space for smoother typing
- On submit, only mentions that were selected from the suggestion list are normalized to file paths (manual mentions like `@alice` are not rewritten)
###  Search + filtering
- Uses `tinyglobby` for fast file discovery
- Honors `.gitignore` via `ignore` (spec-compliant matching)
- Excludes common heavy folders at glob-time (`node_modules`, `.git`, `dist`, etc.)
- Caches file list (5s) and `.gitignore` content (5m)
- Dedupes concurrent scans and limits suggestions to 20 results
 ### What changed
- New: `source/components/file-suggestions/*` (suggestion UI + debounced search)
- Updated: `source/components/input-with-history.tsx` (integration, selection handling, mention rewrite on submit)
- Updated: `source/app.tsx` (submit plumbing)
- Updated: `package.json`, `package-lock.json` (add `ignore`)

 #### Why `ignore` (gitignore parsing) is important
`.gitignore` rules are not plain globs; Git’s matching has special cases (negation `!`, anchoring with `/`, directory-only patterns with trailing `/`, `**` behavior, escaping `#`/`!`, whitespace rules). Using a spec-compliant parser avoids surprising suggestions like ignored build artifacts appearing, or negated “keep” rules not working.

We use the `ignore` library because it implements the `.gitignore` spec (and explicitly notes that common glob matchers like `minimatch` are not spec-compliant). See:
- https://www.npmjs.com/package/ignore
- https://git-scm.com/docs/gitignore

image

Closes #110

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,6 +26,7 @@
         "figures": "^6.1.0",
         "highlight.js": "^11.11.1",
         "html-to-text": "^9.0.5",
+        "ignore": "^7.0.5",
         "ink": "^6.6.0",
         "json-schema": "^0.4.0",
         "json5": "^2.2.3",
@@ -252,6 +253,7 @@
       "integrity": "sha512-e7jT4DxYvIDLk1ZHmU/m/mB19rex9sv0c2ftBtjSBv+kVM/902eh0fINUzD7UwLLNR+jU585GxUJ8/EBfAM5fw==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.5",
@@ -1795,6 +1797,7 @@
       "integrity": "sha512-NMv9ASNARoKksWtsq/SHakpYAYnhBrQgGD8zkLYk/jaK8jUGn08CfEdTRgYhMypUQAfzSP8W6gNLe0q19/t4VA==",
       "devOptional": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@types/node": "*"
       }
@@ -1860,6 +1863,7 @@
       "integrity": "sha512-MWtvHrGZLFttgeEj28VXHxpmwYbor/ATPYbBfSFZEIRK0ecCFLl2Qo55z52Hss+UV9CRN7trSeq1zbgx7YDWWg==",
       "devOptional": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "csstype": "^3.2.2"
       }
@@ -2252,6 +2256,7 @@
       "integrity": "sha512-eGbYq2CT+tos1fBwLQ/tkBt9J5M3JEHjku4hbvQUePCckkvVf14xWj+1m7dGoK81M/fOjFT7yM9UMeKT/+vFLQ==",
       "hasInstallScript": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "bindings": "^1.5.0",
         "prebuild-install": "^7.1.1"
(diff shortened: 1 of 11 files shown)


hazrid93 (commented):
Great work, although i feel like the file search result could use some caching

tanvesh01:
> Great work, althofh i feel like the file searche result could use some caching

Makes sense! I have added a pretty basic version of file caching and respecting gitignore files. Trying to not overcomplicate things for now, as I am sure we are going to build up on this : )

@hazrid93

tanvesh01:
> Using this, it behaves slightly differently than Claude Code's `@` file mentioning. In Claude Code if you `@` mention a file it will always force the model to call the `Read` tool on the mentioned files (at least that's my understanding). Your change seems to just fill in the filename and removes the user's typed `@`.
> 
> It does seem to work well for me, just I noticed this difference in execution. But definitely a handy feature! :)

huh, I really thought just mentioning the filenames was enough to make the LLM read these files. Because thats what happens just now too in this branch for me. But yea good point, maybe I can check how folks at opencode/other-clis do it.

tanvesh01:
@bradfa Do you wanna give it another check please? It should behave very similar to something like claude/open code now. Depending on the file length we give the file's content in the prompt itself

bradfa:
Seems to work nicely now! :)

I like that files which aren't tracked by git but also are not ignored by git show up in the choices.

For some reason my muscle memory wants to press the Tab key to select the current top auto-completed file name instead of pressing Enter.  I'm not married to using Tab to select the choice but that's just what my fingers default to doing for some reason.  Enter works, too and is reliable in my quick tests.

I would suggest you rebase your changes onto the current main branch, possibly squash your commits into a smaller number of changes, and clearly explain what new node package dependencies are being added to octo via this PR.

reissbaker (commented):
Generally speaking I like this PR. However, I think auto-inserting extra `user` messages behind the scenes could cause some... interesting... model behavior in some models, since many were trained assuming a 1:1 relationship between assistant and user messages. I know Claude auto-reads the file, but Anthropic has pretty tight control of how Claude is trained, so we might want to be a little more generic here.

Rather than creating a new type of HistoryItem and compiling down to multiple user messages, IMO it would be simplest to just replace the `@` with the actual path to the file in the `content`, and let the model call its tools normally. I.e. instead of sending:

```
please check @from
```

We'd silently compile down to:

```
please check ./from
```

Sorry for the churn on this one! We were a bit swamped on managing Kimi K2.5 😅

reissbaker (approved):
Neat, LGTM!

PR #114, opened 2026-02-02
Concurrency benchmarking

reissbaker:
Adds a `--concurrency` option to simplify using Octo to simulate high load on GPUs.
image

### DIFF
--- a/source/cli.tsx
+++ b/source/cli.tsx
@@ -216,6 +216,7 @@ bench
     "--prompt <prompt>",
     "Custom prompt to benchmark with. If omitted, uses the default prompt.",
   )
+  .option("--concurrency <n>", "Concurrent requests to make. If omitted, defaults to 1")
   .action(async opts => {
     const { config } = await loadConfigWithoutReauth();
     const model = opts.model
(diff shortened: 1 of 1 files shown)


source/cli.tsx
```diff
@@ -229,81 +230,167 @@ bench
       process.exit(1);
     }
 
+    const concurrency = Math.max(1, parseInt(opts.concurrency ?? "1", 10));
     const apiKey = await assertKeyForModel(model, config);
     const autofixJson = makeAutofixJson(config);
+    const modelToUse = model;
```
sdjidjev, line 236: does this need to be set to a new constant? doesn't seem like `model` is changing anywhere, can you just use that?

sdjidjev (approved):

source/cli.tsx
reissbaker: derp k2.5 dumb

reissbaker: actually I AM dumb, K2.5 had a totally valid reason for this and refactoring it made it annoying. Basically: TypeScript will allow you to narrow a type like so:

```typescript
const x = y; // Assume y is string or undefined
if(x == null) process.exit(1);
// you can now treat x as string
```

However... there's one caveat. TypeScript treats const and let narrowing identically. Let's assume this statement was instead:

```typescript
let x = y;
if(x == null) process.exit(1);
// you can now treat x as a string
```

However! What about this code?

```typescript
let x = y;
if(x == null) process.exit(1);
setTimeout(() => console.log(x.length), 50);
x = undefined;
```

This would of course throw an error at runtime if TypeScript allowed it: `undefined` has no `.length` property. So, TypeScript doesn't allow this code: it only narrows types for the current scope, and it doesn't narrow them in callbacks since it doesn't know when the callbacks will run.

Since TS treats `let` and `const` narrowing identically, inside the callback `runSingleBenchmark`, `model` is possibly-null. There are a variety of workarounds for that, but TBH, the simplest approach is: in a scope where `model` is narrowed to non-nullable, assign it to a new `const`, which will be type-inferred as non-nullable. Then TypeScript knows *that* variable can never be null, and so that's why K2.5 added the extra `const` assignment.

sdjidjev: Woahhh i see, because it's in the callback, TIL 🤔 🤯
Should we then change the `console.log` statement below to use `modelToUse` instead?

reissbaker: Oh, I think it's okay since the `console.log` is in the current scope! So, `tsc` already narrowed the value to non-nullish. It's only in the callback that you need to use the cached non-nullish reference.

PR #117, opened 2026-02-03
Markdown inner-content rendering fix

sdjidjev:
**Before:**
Screenshot 2026-02-02 at 4 25 39 PM
**After:**
Screenshot 2026-02-02 at 5 30 21 PM

- Made `renderChildren` which in turn recursively calls `TokenRenderer`
- Should be no risk of infinite recursion because tokens are processed through `marked.lexer(markdown)` at the beginning

### DIFF
--- a/source/markdown/index.tsx
+++ b/source/markdown/index.tsx
@@ -24,6 +24,10 @@ export function Markdown({ markdown }: { markdown: string }) {
   );
 }
 
+function renderChildren(tokens: Token[]): React.ReactNode {
+  return tokens.map((token, index) => <TokenRenderer key={index} token={token} />);
+}
+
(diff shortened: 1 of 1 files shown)


source/markdown/index.tsx
```diff
@@ -112,7 +116,7 @@ function CodeRenderer({ token }: { token: Tokens.Code }) {
 }
 
 function CodespanRenderer({ token }: { token: Tokens.Codespan }) {
-  return <Text inverse> {token.text} </Text>;
+  return <Text inverse>{token.text}</Text>;
```
sdjidjev, line 119: removed the surrounding spaces because i think it flows better without them

Issue #125, opened 2026-02-06
`npm test` has failures


bradfa:
Current main branch HEAD (a1811e5831b5fda192f9f1648e925a997e166f84) seems to have test failures:

```
$ npm test

> octofriend@0.0.52 test
> vitest


 DEV  v3.2.4 /home/andrew/git/octofriend

 ✓ source/input-history/index.test.ts (4 tests) 23ms
 ✓ source/skills/skills.test.ts (18 tests) 10ms
stdout | training/fix-json/generate-broken-json.test.ts > JSON fuzzing functions > comma manipulation functions > cutComma should cut a comma from the available positions
{
  type: 'array',
  node: {
    type: 'array',
    commaPositions: [ 1, 3, 5 ],
    children: [ [Object], [Object], [Object] ],
    start: 0,
    end: 7,
    length: 7,
    value: [ 1, 2, 3 ]
  },
  broken: {
    commaCuts: Set(1) { undefined },
    commaDupes: Set(0) {},
    openCut: false,
    closeCut: false
  }
}

stdout | training/fix-json/generate-broken-json.test.ts > JSON fuzzing functions > bracket manipulation functions > cutClose should cut closing bracket
{ broken: { openCut: false, closeCut: true } }

 ❯ training/fix-json/generate-broken-json.test.ts (19 tests | 2 failed) 13ms
   ✓ JSON fuzzing functions > null manipulation functions > nullCut should cut a character from null 1ms
   × JSON fuzzing functions > null manipulation functions > nullAdd should insert random characters into null 4ms
     → expected 'nunll' to deeply equal 'numll'
   ✓ JSON fuzzing functions > boolean manipulation functions > boolCut should cut a character from true 0ms
   × JSON fuzzing functions > boolean manipulation functions > boolAdd should insert random characters into boolean 0ms
     → expected 'fanlse' to match 'famlse'
   ✓ JSON fuzzing functions > number manipulation functions > numberDot should add a decimal point 0ms
   ✓ JSON fuzzing functions > string manipulation functions > strUnescape should unescape escaped characters 0ms
   ✓ JSON fuzzing functions > string manipulation functions > strRemoveQuote should remove a quote from string 0ms
   ✓ JSON fuzzing functions > string manipulation functions > strEscapeQuote should escape quotes 0ms
   ✓ JSON fuzzing functions > string manipulation functions > strPrefixQuote should prepend quote 0ms
[...]

PR #126, opened 2026-02-06
Fix broken randomness mock in JSON fuzzing tests

bradfa:
The mockRandom helper used || to provide a fallback value, which caused the mock to return 0.5 instead of 0 when 0 was explicitly specified in the test values array. This happened because 0 is falsy in JavaScript.

When commit de5250b ("Fix randomness functions") corrected the zeroToN function to no longer subtract 1 from its result, the interaction with the buggy mock changed: what previously produced 'm' now produced 'n'.

The tests were likely written to match observed behavior rather than correct behavior, since both the mock and zeroToN were buggy when the tests were introduced. Using ?? instead of || ensures 0 is treated as a valid return value, and the expectations are updated to reflect the actual character produced when the mock returns 0 ('a', not 'm' or 'n').

Fixes: https://github.com/synthetic-lab/octofriend/issues/125

### DIFF
--- a/training/fix-json/generate-broken-json.test.ts
+++ b/training/fix-json/generate-broken-json.test.ts
@@ -26,7 +26,7 @@ describe('JSON fuzzing functions', () => {
   const mockRandom = (returnValues: number[]) => {
     let index = 0;
     return () => {
-      return returnValues[index++] || 0.5;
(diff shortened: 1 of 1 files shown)


sdjidjev (approved):
I've noticed this and wondered why, thanks for the fix!

Issue #128, opened 2026-02-08
Allow reuse of existing API key for new model on existing provider


bradfa:
When adding a new model, if we already have an API key for the provider, then give an option to reuse it instead of having to enter it again.

Since we only map 1 API key per provider URL, it makes sense that we should allow the user to easily reuse the API key which octo already knows.

PR #131, opened 2026-02-08
Allow reusing existing API key when adding model

bradfa:
When adding a new model, if the user already has an API key configured for the model provider (ie: Synthetic, local, Anthropic, etc) then offer to let the user reuse that existing API key.

Since the current mapping of provider to API keys only allows 1 API key per provider (not per model), this is reasonable to do when we already have a provider setup in the config.

Fixes: https://github.com/synthetic-lab/octofriend/issues/128

### DIFF
--- a/source/config.ts
+++ b/source/config.ts
@@ -360,17 +360,7 @@ async function findSyntheticKey(config: Config | null) {
   const override = overrides == null ? null : overrides["synthetic"];
   if (override) return process.env[override];
 
-  // Every API base URL Synthetic has ever used
-  const validBaseUrls = [
-    "https://api.synthetic.new/openai/v1",
-    "https://synthetic.new/api/openai/v1",
-    "https://api.synthetic.new/v1",
-    "https://api.glhf.chat/v1",
-    "https://glhf.chat/api/v1",
-    "https://glhf.chat/api/openai/v1",
-  ];
-
-  for (const base of validBaseUrls) {
+  for (const base of SYNTHETIC_BASE_URLS) {
     const key = await readKeyForBaseUrl(base, config);
     if (key != null) return key;
   }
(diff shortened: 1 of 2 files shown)


bradfa:
This needs some rework...  There's more flows possible around auth than I had realized.  And the keys file associates a given base URL to a key (not a model listing to a key) which makes things a little weird for providers who allow you to have more than 1 API key at a time.

bradfa:
But maybe this brings up a larger question about how a user should change their existing API key in Octo through the menu system?

Granted the current situation of inputting another API key when adding another model isn’t ideal but just having a reuse solution isn’t that much better. Although I do agree that for this flow just reusing it is probably best (without asking). 

But it would be good to also have a way to change a base URL’s API key, too.

bradfa:
@sdjidjev I've rebased onto the main branch and pushed a modified commit which just reuses the key if it already exists for a given base URL.  Did a quick test and seems to be working as expected but since you use the shell auth method can you please just verify that I've not unintentionally broken anything there?

Thanks!

source/config.ts
```diff
@@ -474,6 +464,66 @@ export async function readKeyForBaseUrlResult(
...
+ */
+export async function hasExistingKeyForBaseUrl(
+  baseUrl: string,
+  config: Config | null,
+): Promise<boolean> {
+  if (!config) return false;
+
+  const isSynthetic = SYNTHETIC_BASE_URLS.includes(baseUrl);
+
+  // Check existing models for auth with matching base URL
```
sdjidjev: We already have `readKeyForBaseUrl` that we can reuse here!
something like:
```
const key = await readKeyForBaseUrl(baseUrl, config);
if (key != null) return true;

if (SYNTHETIC_BASE_URLS.includes(baseUrl)) {
  for (const url of SYNTHETIC_BASE_URLS) {
    if (url !== baseUrl) {
      const syntheticKey = await readKeyForBaseUrl(url, config);
      if (syntheticKey != null) return true;
    }
  }
}
return false;
```

sdjidjev (commented):
This will be so awesome!! Just one change and should be good to go. Great work!

source/config.ts
bradfa: Ah! Didn't realize this.  I'll clean that up.

bradfa: Fixed in latest force push.

sdjidjev (approved):
🎉

PR #170, opened 2026-03-13
fix: clear performance measures after streamText to prevent memory leak

Zebastjan:
Fixes Node.js MaxPerformanceEntryBufferExceededWarning by clearing performance measures after each streamText call. The AI SDK telemetry creates performance entries that accumulate without being cleared, eventually exceeding the 1,000,001 entry limit.

### DIFF
--- a/source/compilers/responses.ts
+++ b/source/compilers/responses.ts
@@ -1,6 +1,7 @@
 import { createOpenAI } from "@ai-sdk/openai";
 import { streamText, tool, ModelMessage, jsonSchema } from "ai";
 import { t, toJSONSchema } from "structural";
+import { performance } from "node:perf_hooks";
 import { Compiler } from "./compiler-interface.ts";
 import { LlmIR, ToolCallRequest, AssistantMessage } from "../ir/llm-ir.ts";
 import { tryexpr } from "../tryexpr.ts";
@@ -424,6 +425,9 @@ export const runResponsesAgent: Compiler = async ({
           break;
       }
     }
+
+    // Clear performance measures to prevent memory leak from accumulating entries
+    performance.clearMeasures();
 
     // Track usage
     if (usage.input !== 0 || usage.output !== 0) {


sdjidjev:
Cool! 
@Zebastjan was there a prompt you would do that would trigger the `MaxPerformanceEntryBufferExceededWarning` error? or would this happen after a long session?

Zebastjan:
This happened after a long session. I don't recall the particular prompt. If I remember correctly, it was albe to recover from the error.

source/cli.tsx
```diff
@@ -112,9 +113,26 @@ docker
...
+    configPath = loaded.configPath;
+  } catch (error) {
+    // Handle config loading errors gracefully
+    console.error("\n❌ Failed to load configuration\n");
+    if (error instanceof Error) {
+      console.error(error.message);
+    } else {
+      console.error(String(error));
+    }
+    console.error("\nRun 'octofriend init' to create a fresh configuration.\n");
```
reissbaker: This `try/catch` block looks overly LLM-y to me:

1. `loadConfig` will already do first-time setup if necessary: running `octofriend init` isn't the right solution for errors here.
2. It's not clear to me what errors here are attempting to be caught, or why.

IMO this is unrelated to performance metrics and should probably be removed from this PR.

reissbaker (changes requested):
Perf metrics fix looks fine!

The extra defensive `try/catch` in `runMain` should probably be removed though

Zebastjan:
Thanks for the feedback. I'll take a look at it.

Zebastjan:
Removed the extra try/catch in runMain as requested. The PR now only contains the performance fix.

sdjidjev (approved):

sdjidjev:
Sorry accidentally pushed a merge commit I did, reset!

sdjidjev:
I'll take a look to see if this issue still exists with `octo` running on paintcannon now. If so, I'll open up a new PR crediting you with your fix. Closing this PR right now to clean up the PR list!

PR #172, opened 2026-03-15
feat: batch pseudo-tool for parallel/sequential multi-tool execution

Zebastjan:
## Problem

Tasks requiring codebase exploration or multi-resource fetching generate N sequential tool calls — one LLM round-trip per call. This inflates inference cost, consumes context window budget, and produces repetitive tool-call noise that degrades reasoning quality.

## Solution

A `batch` pseudo-tool that accepts a declarative list of tool calls and executes them in a single LLM turn, returning TOON-formatted results. TOON's tabular format reduces token count ~40% vs equivalent JSON tool results, and the LLM reasons more cleanly over a single structured block than N fragmented tool messages.

## Usage

The LLM calls batch with an array of sub-calls and an optional parallel flag. Results come back as a pipe-delimited TOON table:

```
@results[tool,success,content,error]
read|true|<file contents>|
read|true|<file contents>|
```

## Design Notes

- **Execution:** sequential by default, parallel via `Promise.allSettled()`
- **Error handling:** per-sub-call, only fails outer tool if ALL fail
- **AbortSignal** propagated to all sub-tools
- **Confirmation** delegated to sub-tools (batch itself is auto-approved)
- **No new dependencies** (inline TOON serializer, ~10 lines)

## Testing

Tested locally with Octofriend v0.0.53 + Kimi K2.5:
- Sequential batch of 3 files: correct ordering preserved, all results
  returned in single TOON table (4 lines: header + 3 rows)
- Parallel batch of 3 files with 1 non-existent path: 2 succeeded,
  1 failed with "No such file" error in error column — batch did not
  crash, remaining results unaffected

## Future Direction

This PR lays the groundwork for a local helper model integration (following Octofriend's existing diffApply/fixJson pattern) that would synthesize batch results locally before they reach the backend LLM — further reducing inference noise and context window consumption. Synthetic already ships fine-tuned local models for diff and JSON repair; a synthesis model is the natural third member of that family.

### DIFF
--- a/BATCH-TOOL-INSERTION-POINT.md
+++ b/BATCH-TOOL-INSERTION-POINT.md
@@ -1,0 +1,354 @@
+# BatchToolExecution Insertion Point Analysis
+
+## Overview
+
+This document identifies the exact files and functions where a `BatchToolExecution` pseudo-tool could be integrated into the Octofriend codebase.
+
+## Tool Registration Point
+
+### Option 1: New Tool Module (Recommended)
+
+**File to create:** `source/tools/tool-defs/batch.ts`
+
+**Registration:** Add to `source/tools/tool-defs/index.ts:15-29`
+
+```typescript
+import batch from "./batch.ts";
+
+export default {
+  read,
+  list,
+  // ... other tools
+  batch, // <-- Add here
+};
+```
+
+**Pattern to follow (from `source/tools/tool-defs/read.ts:1-49`):**
+
+```typescript
+import { t } from "structural";
+import { defineTool } from "../common.ts";
+
+const ArgumentsSchema = t.subtype({
+  calls: t.array(
+    t.subtype({
+      tool: t.str,
+      arguments: t.any,
+    }),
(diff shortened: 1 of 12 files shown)


reissbaker:
While I agree we should support batched tool calls, we shouldn't invent a new batch format: instead we should use the built-in parallel tool call support of the various OpenAI/Anthropic-compatible APIs. This would involve updating the LlmIR and HistoryItems to support multiple tool calls returned by an assistant, and would require updating the various compilers to include support for that.

We also shouldn't use alternative data formats (e.g. TOON), since LLMs are trained on their native tool calling format and using alternatives will push them out of distribution.

Zebastjan:
Thanks for the feedback and the pointers on how you’d like this structured — that’s really helpful. We’ll rework this to focus just on the batch tooling and then resubmit the PR with that scope.TOON (Token-Oriented Object Notation) support will be handled as a separate issue so it doesn’t complicate this change.

PR #173, opened 2026-03-17
fix: reconnect MCP clients after server restart

Zebastjan:
## Problem

When an MCP server process is killed and restarted (e.g., due to a crash or manual restart), the cached `Client` object becomes stale and throws `Error: Not connected` on subsequent calls.

## Solution

Add a health check in `getMcpClient()` that verifies cached connections with a lightweight `listTools()` call before returning them. If the connection is dead, we:
1. Remove the stale entry from the cache
2. Create a fresh connection

## Changes

- `source/tools/tool-defs/mcp.ts`: +8 lines

## Testing

- ✅ Build passes
- ✅ All existing tests pass (112 tests)
- ✅ Manual verification: stale connections are detected and reconnected automatically
- ✅ Health check overhead: ~27ms (acceptable)

## Reproducing the Bug

1. Start an MCP server (e.g., Archon)
2. Use an MCP tool (caches the client)
3. Kill the MCP server process
4. Restart the server on the same port
5. Try to use an MCP tool again → `Error: Not connected`

With this fix, step 5 automatically reconnects instead of crashing.

### DIFF
--- a/BATCH-TOOL-INSERTION-POINT.md
+++ b/BATCH-TOOL-INSERTION-POINT.md
@@ -1,0 +1,354 @@
+# BatchToolExecution Insertion Point Analysis
+
+## Overview
+
+This document identifies the exact files and functions where a `BatchToolExecution` pseudo-tool could be integrated into the Octofriend codebase.
+
+## Tool Registration Point
+
+### Option 1: New Tool Module (Recommended)
+
+**File to create:** `source/tools/tool-defs/batch.ts`
+
+**Registration:** Add to `source/tools/tool-defs/index.ts:15-29`
+
+```typescript
+import batch from "./batch.ts";
+
+export default {
+  read,
+  list,
+  // ... other tools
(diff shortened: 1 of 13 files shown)


Zebastjan:
This could also potentially be fixed by a lazy retry, but I don't know. Let me know how you like this and if not, we can redo it the other way.

sdjidjev:
hmm looks like this PR contains multiple unrelated commits
@Zebastjan could you make this PR just contain the commit about the MCP client reconnection?

Zebastjan:
> hmm looks like this PR contains multiple unrelated commits @Zebastjan could you make this PR just contain the commit about the MCP client reconnection?

Yes, sorry about that!

Zebastjan:
Closing in favor of #178 - this PR was contaminated with unrelated commits (config validation improvements and batch tool feature). The clean PR contains only the MCP reconnection fix.

PR #178, opened 2026-03-23
fix: reconnect MCP clients after server restart

Zebastjan:
## Problem

When an MCP server process is killed and restarted (e.g., due to a crash or manual restart), the cached Client object becomes stale and throws `Error: Not connected` on subsequent calls.

## Solution

Add a health check in `getMcpClient()` that verifies cached connections with a lightweight `listTools()` call before returning them. If the connection is dead, we:

1. Remove the stale entry from the cache
2. Create a fresh connection

## Changes

- `source/tools/tool-defs/mcp.ts`: +9 lines, -1 line

## Testing

- ✅ Build passes
- ✅ All existing tests pass
- ✅ Manual verification: stale connections are detected and reconnected automatically
- ✅ Health check overhead: ~27ms (acceptable)

## Reproducing the Bug

1. Start an MCP server (e.g., Archon)
2. Use an MCP tool (caches the client)
3. Kill the MCP server process
4. Restart the server on the same port
5. Try to use an MCP tool again → Error: Not connected

With this fix, step 5 automatically reconnects instead of crashing.

---

**Note**: This is a clean replacement for #173, which was contaminated with unrelated commits (config validation improvements and batch tool feature).

### DIFF
--- a/source/tools/tool-defs/mcp.ts
+++ b/source/tools/tool-defs/mcp.ts
@@ -66,7 +66,15 @@ const clientCache = new Map<string, Client>();
 export async function getMcpClient(serverName: string, config: Config): Promise<Client> {
   // Check cache first
   if (clientCache.has(serverName)) {
-    return clientCache.get(serverName)!;
+    const cachedClient = clientCache.get(serverName)!;
+    // Try a lightweight operation to verify connection is still alive
+    try {
+      await cachedClient.listTools();
+      return cachedClient;
+    } catch {
+      // Connection lost, remove from cache and reconnect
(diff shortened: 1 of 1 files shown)


Zebastjan:
Unfortunately, we've had problems with pull requests that have been contaminated by unrelated  work.  We are making changes to our workflow to prevent the problem from reoccurring in the future.  Our apologies.  Please bear with us as we get this all sorted out properly so that in future our pull requests  are clean and focused.

sdjidjev:
Closing for now to clean up PR list!

PR #179, opened 2026-03-24
Make cwd a string property in Transport

sdjidjev:
Separated this out into its own PR since it was required for LSP to work without relying on an async cwd() function. 

Instead of having `cwd` be a function in `transport`, it's now a `string` which gets initialized on app start.

### DIFF
--- a/source/transports/docker.ts
+++ b/source/transports/docker.ts
@@ -50,24 +50,42 @@ type DockerTarget =
       type: "image";
       image: Awaited<ReturnType<typeof manageContainer>>;
     };
+
 export class DockerTransport implements Transport {
   private readonly _container: string;
+  cwd: string;
 
-  constructor(private readonly _target: DockerTarget) {
+  private constructor(
+    private readonly _target: DockerTarget,
+    cwd: string,
+  ) {
     if (this._target.type === "image") this._container = this._target.image.container;
     else this._container = this._target.container;
+    this.cwd = cwd;
+  }
+
+  static async create(target: DockerTarget): Promise<DockerTransport> {
+    const container = target.type === "image" ? target.image.container : target.container;
+    const cwd = await DockerTransport.runDockerCommand(
+      container,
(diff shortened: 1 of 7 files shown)


source/transports/docker.ts
```diff
@@ -143,17 +161,22 @@ output: ${output}`,
...
+
   async writeFile(signal: AbortSignal, file: string, contents: string): Promise<void> {
     // Create a temporary file with the contents
     const tempFile = `/tmp/octo_write_${randomSuffix()}`;
 
     // First, write the contents to the temp file using a base64 to avoid shellescape issues
     const base64Contents = Buffer.from(contents).toString("base64");
-    await this.dockerExec(
-      signal,
-      ["/bin/sh", "-c", `echo '${base64Contents}' | base64 -d > '${tempFile}'`],
```
sdjidjev, line 154: seems like this used to be a bug? where "/bin/sh" and "-c" would be in the command twice

source/transports/docker.ts
```diff
@@ -50,24 +50,42 @@ type DockerTarget =
...
+    return new DockerTransport(target, cwd.trim());
   }
 
   async close() {
     if (this._target.type === "image") await this._target.image.close();
   }
 
-  private async dockerExec(
-    signal: AbortSignal,
+  private static runDockerCommand(
```
reissbaker: Nit: rather than using `private static` functions, I think it's simpler to just use regular functions that you don't export:

```typescript
async function runDockerCommand(
  container: string,
  // ...
): Promise<string> {
  // ...
}
```

Since you're not accessing anything on the `DockerTransport` class, it doesn't need to be a method: it can just be a function. It also shortens callsites, i.e. `runDockerCommand(...)` rather than `DockerTransport.runDockerCommand(...)`!

reissbaker (approved):
Seems reasonable! Left a tiny nit but otherwise LGTM :)

PR #182, opened 2026-04-02
Add LSP tool

sdjidjev:
[Official LSP Spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/)

Adds LSP (Language Server Protocol) functionality to Octo!

**Core LSP Infrastructure:**
- Full LSP client implementation (`LspClient`) that communicates with LSP servers via stdio
- Automatic LSP server discovery and detection for file extensions
- Project-based LSP configuration via `.octofriend/lsp.json5`
- LSP server recommendations based on file type
- Caching of LSP clients across operations
- File version tracking for consistent LSP state

**Available LSP Tools:**
- `lsp-definition` - Find where a symbol was originally defined
- `lsp-references` - Find all references to a symbol
- `lsp-hover` - Get type info and documentation for a symbol
- `lsp-diagnostics` - Get errors and warnings for a file
- `lsp-implementation` - Find implementations that jump past interfaces/abstract classes
- `lsp-document-symbol` - List all symbols (functions, classes, variables) in a file
- `lsp-incoming-calls` - Find all callers of a function
- `lsp-outgoing-calls` - Find all functions called by a function

**Configuration:**
- Global LSP enable/disable via `config.lsp`
- Per-server disable options (e.g., `config.lsp.rust-analyzer.disabled`)
- Automatic server detection and startup based on available executables
During a LSP tool call, if a recommended LSP server is found to perform an LSP action for a file, but the user doesn't have it installed yet, we show a LspRecommendationPrompt to let Octo auto-install the LSP and continue operations:
Screenshot 2026-04-02 at 4 08 30 PM

_(in unchained mode, the LSP will auto install using the shell command from the list of pre-approved LSP recommendations)_

I commented by hand a lot of the LSP specific behavior since some of it is not general knowledge and there are some nuances.

To disable a specific LSP server recommendation, in `~/.config/octofriend/.octofriend.json5` you can do:
```
  lsp: {
    'typescript-language-server': {
      disabled: true,
    },
  }
```
or if you want all LSP recommendations disabled:
```
lsp: false,
```

Users can define their own LSP server configs in their project's directory in a file in `.octofriend/lsp.json5`
Example:
```
{
  'typescript-language-server': {
    serverName: 'typescript-language-server',
    command: [
      'typescript-language-server',
      '--stdio',
    ],
    extensions: [
      '.ts',
      '.tsx',
      '.js',
      '.jsx',
      '.mjs',
      '.cjs',
      '.mts',
      '.cts',
    ],
    rootCandidates: [
      'package.json',
      'tsconfig.json',
      'jsconfig.json',
    ],
  },
}
```

Screenshots of some LSP actions at work!
Screenshot 2026-04-02 at 4 25 21 PM
Screenshot 2026-04-02 at 4 51 53 PM
Screenshot 2026-04-02 at 4 51 45 PM
Screenshot 2026-04-02 at 4 51 37 PM
Screenshot 2026-04-02 at 4 51 31 PM

\* \- Octofriend should probably start having more detailed documentation on what LSP actions are supported and how to configure the setup locally (i'm inspired by how [opencode does theirs](https://opencode.ai/docs/lsp/)), so will definitely be a follow-up task.

PS - I know this PR is super big, idk if it would be easier to review if it was chopped up into chunks but I think one of y'all mentioned you like each PR should include the whole feature so I'll leave it as-is. Since it was already so big, I left out some features (like initializationOptions, workspaceSymbols, LSPs that can't be installed with 1 shell command, etc.) but can add those as a follow-up!
That said, lmk if you need any clarification or PR separating!!


source/prompts/system-prompt.ts
```diff
@@ -124,6 +124,30 @@ have large test suites that take a very long time to run.
 
 IMPORTANT: DO NOT ADD ANY COMMENTS unless asked
 
+# LSP (Language Server Protocol)
```
reissbaker: You don't need to add this to the system prompt! A better place to add it is to the LSP tool itself, using Structural's `.comment(...)` function. Structural automatically converts `.comment(...)` calls into JSON Schema descriptions when using `toJSONSchema(...)`, so that way the instructions will exist if the tool exists and will be skipped otherwise! LLMs might also respect the JSON Schema descriptions with tool defs more than just pure system prompts, although that will depend on the model.

source/tools/tool-defs/lsp.ts
```diff
@@ -0,0 +1,218 @@
...
+    ),
+});
+
+const Schema = t
+  .subtype({
+    name: t.value("lsp"),
+    arguments: ArgumentsSchema,
+  })
+  .comment(
+    "Query a running Language Server for code intelligence. Use this tool INSTEAD of reading files when you need type info, definitions, references, or symbol lookups since it is faster and more accurate than reading and searching manually. Inspect the action parameter's options and their descriptions to determine which action to use.",
```
reissbaker: IMO we should make it clear to an LLM what languages we have LSP servers configured for, so that they don't bother calling the tool if they're working in a language we don't have a definition for. You can do this by calling `.comment(...)` in the `defineTool` function, so that you have full access to the config file and can tell what LSP servers are options for the LLM. The LSP file types are probably just:

- The built-in ones, and
- Whatever is defined in the config.json5

source/state.ts
```diff
@@ -26,6 +26,8 @@ export type RunArgs = {
   transport: Transport;
 };
 
+export type InstallationChoice = "install" | "skip" | "never" | "disable-all";
```
reissbaker: Since this isn't used inside `state.ts`, should this be defined in a different file (i.e. where it's used)?

source/tools/tool-defs/lsp.ts
```diff
@@ -0,0 +1,218 @@
...
+      t
+        .value("references")
+        .comment("find all references to a symbol. Requires file, line, character"),
+    )
+    .or(
+      t
+        .value("hover")
+        .comment("type info and documentation for a symbol. Requires file, line, character"),
+    )
+    .or(t.value("diagnostics").comment("errors and warnings for a file. Requires file only"))
```
reissbaker: I think you can encode the types and requirements more elegantly, and in a way that's less likely to result in the LLM making mistakes and also leans on Structural for validation. It looks like you need different arguments for different subcommands; there are basically two ways to do this:

1. Encode this in each action, or
2. Make multiple tools for each action

(2) sounds wasteful, *but* — originally, file editing was done like (1), but I saw some open-source models were kinda too dumb to figure out complex schemas, and breaking it out into multiple tools worked better.

I think it's worth trying (1) for elegance, since it's closer to your current implementation anyway. But, it may be the case we eventually denormalize all of these actions into separate tools, if it helps the models be smarter. Anyway, here's how I'd do (1):

```typescript
const LineSchema = t.num.comment(
  "1-indexed line number"
);
const CharSchema = t.num.comment(
  "1-indexed column number"
);
const DefinitionSchema = t.subtype({
  action: t.value("definition"),
  line: LineSchema,
  character: CharSchema,
});
const ReferencesSchema = t.subtype({
  action: t.value("references"),
  line: LineSchema,
  character: CharSchema,
});

// ...etc

const ArgumentSchema = t.subtype({
  filePath: t.str,
  action: DefinitionSchema.or(
    ReferencesSchema
  ).or(
    // ...
  ),
});
```

That way, there's a few advantages:

1. Structural can automatically guarantee the type is correct without runtime checks
2. The LLM can see the exact type it needs without needing to jump around mentally from comments (unclear whether this is valuable, but maybe?)
3. Inference engines that support "strict" function calling, where the LLM output is constrained based on a JSON schema passed in, will guarantee that the LLM never makes a mistake on this call. (Sadly we don't support strict calling on all models, but we do on some at least)

During this refactor, you may be tempted to extract out a shared definition like:

```typescript
const RequiresAll = t.subtype({
  line: LineSchema,
  character: CharSchema,
});

const DefinitionSchema = RequiresAll.and(t.subtype({
  action: t.value("definition"),
}));
```

However — this actually won't work for many inference engines. Sadly many inference engines are written in Python by monkeys, and so they don't support exotic features such as [JSON Schema allOf](https://json-schema.org/understanding-json-schema/reference/combining), which is what `.and(...)` compiles down to. Instead, if you want to do that, you'll need to denormalize it in TypeScript using splats:

```typescript
const requiresAll = {
  line: LineSchema,
  character: CharSchema,
} as const;

const DefinitionSchema = t.subtype({
  action: t.value("definition"),
  ...requiresAll,
});
```

source/app.tsx
```diff
@@ -781,10 +820,20 @@ function ToolRequestRenderer({
...
   useEffect(() => {
-    if (noConfirmationNeeded) {
+    const noLspInstallationNeeded = toolFn.name !== "lsp" || !lspMayNeedInstall;
+    if (noConfirmationNeeded && noLspInstallationNeeded) {
       runTool({ toolReq, config, transport });
     }
-  }, [toolReq, noConfirmationNeeded, config, transport]);
+  }, [toolReq, noConfirmationNeeded, config, transport, runTool, lspMayNeedInstall]);
+
+  if (toolFn.name === "lsp" && lspRecommendation) {
```
reissbaker: I might not be following the flow of the code correctly, but this feels a little flash-of-content-y to me, since `lspRecommendation` is set by an async function inside of a `useEffect`. Do we need a `loading` state, to render some interstitial "loading LSP..." text while we decide which thing to render?

reissbaker (commented):
Approach generally looks good to me! Left some comments on the tool impl + a React component question.

One note: when I use Octo and it gets (for example) compiler output from `tsc`, I've noticed it often can't actually tell what line numbers map to what lines. For example, if it sees there's an error on line 101, it might not correctly figure out that means:

```
const x = Math.floor(); // haha no input, error
```

because we don't ever show it numbered lines; i.e. we'll send it an entire file, and it has to essentially internally count the lines itself.

I've noticed that Claude often internally compiles files down to lines with actual line numbers; for example, something like:

```
filename: README.md
1. # README.md
2. Hello, this is the
3. README for the
4. tool Octofriend
```

Have you noticed issues with Octo's ability to call different LSP tools accurately, since it's not great at guessing line/col numbers and some of these tools require specific line or column numbers?

I wonder if we should start compiling down file output to include line numbers in the compilers, so that Octo can use line-number-based tooling more effectively.

I also wonder about the character/column line numbers from the perspective of "can an LLM correctly use this most of the time": I would guess many LLMs are bad at counting characters in a line. If so, it might be worth exposing a different API for the column tooling, e.g. a string to search for in the line (and then on our end, use `.indexOf(searchString) + 1` before passing to the LSP server).

source/tools/tool-defs/lsp.ts
sdjidjev: @reissbaker - thanks for the comments! However ran into something trying to implement suggestion (1).

I definitely prefer the code elegance of option (1), but I ran into some issues where for example GLM4.7 was passing back a tool request where it was double-stringifying (because of the nested objects)
If I change octo's compiler to recursively decode stringified tool arguments it does work, but wondering if we should avoid that instead and do option (2) instead.

When I was first building this out, I thought about doing option (2) and having 1 tool per LSP action but was told by AI that Tool Saturation could maybe(?) be an issue and also I saw that OpenCode just has one `LSP` tool for all actions so decided to stick to how they were doing it.

However I think that having lots of `action`s within a LSP tool can't be that much different from having several LSP `tool`s so if (2) sounds better to you I can do it that way instead! Or if you think it's better to recursively decode tool arguments so we can stick with option (1) I'm also down for that!

source/app.tsx
sdjidjev: Ah I could see how it smells like such but it's the same flow as how whitelist is done in this function!
That is to say, there is no flash of content and the prompt only shows up if it's needed.

source/tools/tool-defs/lsp.ts
reissbaker: Let's go ahead and do 2 then!

source/app.tsx
reissbaker: Ah, sounds good!

sdjidjev (commented):
@reissbaker Thanks for the comments! Applied feedback and ready for another review!

Also agree that we should probably pass in encoded lines like Claude does!
Though luckily, for most LSP actions, it doesn't matter as much since the LSP server returns rich information (not just positions), but definitely would make it even more powerful if we numbered files for future processing.
Can do that in a fast-follow PR since this one is already quite large 😅

source/lsp/lsp-server-registry.ts
```diff
@@ -0,0 +1,231 @@
...
+  installCmd: ["npm", "install", "-g", "bash-language-server"],
+};
+
+export const LuaLS: LspInstallationConfig = {
+  serverName: "lua-ls",
+  command: ["lua-language-server"],
+  extensions: [".lua"],
+  rootCandidates: [".luarc.json", ".luarc.jsonc", ".luacheckrc", ".stylua.toml"],
+  description: "Lua language server",
+  installCmd: ["brew", "install", "lua-language-server"],
```
reissbaker, line 54: Oooh — we shouldn't expect that `brew` exists on users' machines; for example, on Linux it's probably unlikely. It's worth highlighting some kind of notification when Octo reads a file if `lua-ls` doesn't exist that Octo *could* use an LSP server if one were installed.

For language-specific servers I think it's safe to use *that language's package manager* (if it exists on the system) to auto-install; for example, using `npm` to auto-install JS/TS language servers, `cargo` for Rust, `go` for Go, etc. But for cases where there isn't a language-specific package manager we should just let the user know they could install the language server if they want to rather than hoping that they have a specific package manager set up.

source/lsp/detect.ts
```diff
@@ -0,0 +1,284 @@
...
+const ProjectLspConfigSchema = t.dict(LspServerConfigSchema);
+
+function getProjectLspConfigPath(cwd: string): string {
+  return path.join(cwd, ".octofriend", "lsp.json5");
+}
+
+export function getUsableLspExtensions(appConfig: Config): Set<string> {
+  const extensions = new Set<string>();
+
+  for (const server of RecommendedLspServers) {
```
reissbaker: Since not all recommended LSP servers are auto-installable (i.e. ones that currently depend on Homebrew, etc), this should probably first check that the LSP server is either:

1. Already installed, or
2. Can be auto-installed

If not we should hide the extension. However, if Octo reads a file from a recommended LSP server extension that fails both of those, we should automatically notify them that if they installed the LSP server we could automatically have Octo use it.

In the future it would be neat to run a sub-agent to auto-install the server, i.e. show a prompt that says something like "Have Octo attempt to install this for you" and it spawns a subagent whose task is to install the LSP server (since Octo could do things like inspect the system, check if Homebrew is installed, check if npm is installed, search default Linux package managers, etc).

reissbaker (commented):
This generally looks good to me, with the caveat that I think auto-installation is probably not safe for some of the LSP servers. For example, on Linux, many people won't have Homebrew installed, since most people will be using their distro's package manager rather than one that ~mostly targets Mac users (technically it's possible to run Homebrew on Linux, but many Linux users don't since they have system-level package managers built-in). Similarly, although technically you need `npm` installed to run Octo, in the future it would be nice to remove this requirement by bundling a fat binary with Bun and distributing it via a shell script (which also simplifies auto-updates), which means we might not reasonably expect that `npm` exists for e.g. Lua programmers.

Left some more detailed comments about it below, but TLDR I think for those cases we should just show people a helpful message that if they *did* install the LSP server, Octo could use it! We could give an example command to run although it may not work for all users (but they can figure the right one for themselves).

sdjidjev:
Ah, derp, excellent catch 😅 Definitely makes sense RE: auto-installation. Slipped my mind since I was mostly testing with `npm`/`javascript` stuff.
It'll make this PR smaller too if we don't include the auto-install, but agree that we should later introduce subagents that could do the installation.
For now I'll change this so we notify the user that installing a certain LSP server will allow `octo` to be more efficient/use less tokens/etc.

sdjidjev:
@reissbaker been thinking about how to do permissions for LSP.

Let's say `lsp-server-registry.ts` has a LSP Server entry called `Sample LSP Server`, which is runnable via `sample-lsp-server-cmd`.

We'll use `which(` to determine whether the command is runnable on the Person's computer.

----- 

Case A:
Person A has `sample-lsp-server-cmd` runnable. Because `Sample LSP Server` is in `lsp-server-registry.ts` we will run the command for LSP tool calls. **(without asking for permission)**

Case B:
Person B has `custom-lsp-cmd` runnable. If they add the appropriate `Custom LSP Server` into their `lsp.json5` file, we will run the command for LSP tool calls. **(without asking for permission)**

Case C:
Person C has `not-included-lsp-cmd` runnable. `Not Included LSP Server` doesn't exist in `lsp-server-registry.ts` nor in their `lsp.json5`, so we never run the command.

Case D:
Person D has `sample-lsp-server-cmd` runnable but they've disabled this LSP server in their `.octofriend.json5` config. We will not run the `sample-lsp-server-cmd` command in any situation.

Case E:
Person E doesn't have any LSP server command installed yet but they haven't disabled any either. For this PR we won't prompt them to install an LSP server. But in a future PR we can prompt them to launch a subagent to attempt installation for them.

----- 

This is what I had in mind for this PR. I wanted to make sure seems stable/secure enough to you to do it in this way and not ask for permission in Cases A & B. _Technically_ if there's any malicious modification to `octo`s config or source code, this could lead to arbitrary code execution via fake LSP commands. But feels edge-casey enough it might not be worth worrying about.

reissbaker:
Yes! That sounds great :)

FWIW, I think it's fine to continue to prompt them to install some LSP servers (i.e. the Go one if they have the `go` binary, which we can check via `which`; the Rust one if they have `cargo` installed; etc). We just prob shouldn't depend on third party ones like `brew`, or expect certain ones to exist e.g. using `npm` to install Lua LSP.

sdjidjev (commented):
@reissbaker thanks for your feedback!

We no longer prompt the user to allow octo to auto-install any LSPs, but if we detect they have one installed that's either in Octo's LSP Server Registry or in their lsp.json5 file, we'll go ahead and use that server. 

However, I kept the `installCmd` data in our server registry to allow for us to suggest a subagent to install it for the user in the future!

Also we dynamically enable/disable the LSP tools based on whether the user has an installed LSP server.

source/lsp/detect.ts
```diff
@@ -0,0 +1,192 @@
...
+
+/**
+ * @returns    set of file extensions that have an installed non-disabled LSP server
+ */
+export async function getUsableLspExtensions(cwd: string, config: Config): Promise<Set<string>> {
+  const usableLsps = await ensureUsableLspsPopulated(cwd, config);
+  return new Set(Object.keys(usableLsps));
+}
+
+// Parses lsp.json5 and populates cachedProjectLspConfig with non-disabled LSPs
```
reissbaker: Any reason to have a separate `lsp.json5` rather than sticking this in the `octofriend.json5` main config? I like having One Big Config, since it's easy to track in dotfiles-style version control. The only other config file we have is the secrets file, but we intentionally keep that one separate since generally you don't want to accidentally check secrets into source control (which you could easily do if it was part of the main `octofriend.json5`)

source/lsp/client.ts
```diff
@@ -0,0 +1,611 @@
...
+      this.latestDiagnostics.set(uri, diagnostics);
+      this.diagnosticsVersion += 1;
+    }
+  }
+}
+
+function fileUri(filePath: string): string {
+  return `file://${path.resolve(filePath)}`;
+}
+export const EXTENSION_TO_LANGUAGE: Record<string, string | undefined> = {
```
reissbaker, line 422: Not necessary for this PR, but it feels like there's a good argument to allow this to be overridable/extensible in a future PR, since AFAICT this static hardcoded list means we can't support LSP tools for languages that aren't in this list?

reissbaker (approved):
Generally LGTM although I think the config should go in the `octofriend.json5` rather than a new `lsp.json5`, unless there's a strong reason for keeping them separate

source/lsp/detect.ts
sdjidjev: Oh I added it when we had the conversation of having project-specific configs that can override the global one. But agree that maybe it's better to just keep it in the main config since the migration code is there too.

source/lsp/client.ts
sdjidjev: oh yes this is an excellent point, 100% agree

PR #183, opened 2026-04-12
Allow configuring a setting that runs after a prompt completes

billycao:
Adds a `notifyFinishCommand` config setting that allows specifying a command that runs after a prompt is complete and octo is waiting for input.

Useful for getting notified after long prompts.

This impl only notifies on finish, not every time Octo requires input (eg. for tool confirmations).

Was inspired to add this feature since I recently switched to a ghostty fork that listens for [OSC sequences](https://ghostty.org/docs/vt/concepts/sequences#osc) to present notifications. 

https://github.com/user-attachments/assets/db6b4fb9-83e4-437a-9b52-a7a7febba017

### DIFF
--- a/README.md
+++ b/README.md
@@ -260,8 +260,12 @@ OCTO_VERBOSE=1 octofriend
 ## Notify Finish Command
 
 You can configure a command to run whenever Octo finishes responding to a prompt.
-This is useful for desktop notifications, sounds, or other integrations. Add a
-`notifyFinishCommand` to your `~/.config/octofriend/octofriend.json5`:
+This is useful for desktop notifications, sounds, or other integrations.
+
+This command will **not** run for tool call confirmations. It will only run when
+Octo is ready for a new prompt (eg. after responding to a prompt in unchained mode)
+
+Add a `notifyFinishCommand` to your `~/.config/octofriend/octofriend.json5`:
 
 ```json5
 notifyFinishCommand: "notify-send Octo 'Finished responding!'",
--- a/README.md
+++ b/README.md
@@ -257,6 +257,22 @@ underlying error messages from APIs or tool calls, run Octo with the
 OCTO_VERBOSE=1 octofriend
 ```
 
+## Notify Command
+
+You can configure a command to run whenever Octo finishes responding to a prompt.
+This is useful for desktop notifications, sounds, or other integrations. Add a
+`notifyCommand` to your `~/.config/octofriend/octofriend.json5`:
+
+```json5
+notifyCommand: "notify-send Octo 'Finished responding!'",
+```
+
+Or for macOS:
+
+```json5
+notifyCommand: 'osascript -e \'display notification "Octo finished!"\'',
+```
+
 ## Opting into canary versions
 
 If you want to use unreleased versions of Octo, clone this repo and read the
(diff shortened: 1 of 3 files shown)


source/config.ts
```diff
@@ -142,6 +144,32 @@ const authCommandCache = new Map<string, string>();
...
+const NOTIFY_COMMAND_TIMEOUT_MS = 15_000;
+
+export async function runNotifyCommand(config: Config): Promise<void> {
+  const cmd = config.notifyFinishCommand;
+  if (!cmd || cmd.trim() === "") return;
+  const shell = process.env["SHELL"] || "/bin/sh";
+
+  await new Promise<void>((resolve, reject) => {
+    const child = spawn(shell, ["-c", cmd], {
+      stdio: ["ignore", "pipe", "pipe"],
```
sdjidjev: it doesn't look like the stdout/stderr streams are read so we could just change to `"ignore"` for all of them to ensure we don't block any threads.

source/config.ts
```diff
@@ -8,6 +8,7 @@ import json5 from "json5";
 import { execFile } from "child_process";
 import { fileExists } from "./fs-utils.ts";
 import { providerForBaseUrl, keyFromName, ProviderConfig } from "./providers.ts";
+import { spawn } from "child_process";
```
sdjidjev: -nit- combine with existing `"child_process"` import

source/config.ts
```diff
@@ -142,6 +144,32 @@ const authCommandCache = new Map<string, string>();
 const AUTH_COMMAND_TIMEOUT_MS = 15_000;
 const AUTH_COMMAND_MAX_OUTPUT_BYTES = 16 * 1024;
 
+const NOTIFY_COMMAND_TIMEOUT_MS = 15_000;
```
sdjidjev: -question- 15 seconds seems a little long, are there notification processes that might take around this long?

README.md
```diff
@@ -257,6 +257,22 @@ underlying error messages from APIs or tool calls, run Octo with the
 OCTO_VERBOSE=1 octofriend
 ```
 
+## Notify Finish Command
+
+You can configure a command to run whenever Octo finishes responding to a prompt.
```
sdjidjev, line 262: I would add that this does not encompass things like requesting permissions, etc.
Or like how this would be true if running in unchained mode.

sdjidjev (approved):
LGTM!  just left some small feedback

source/config.ts
billycao: Done, good catch!

source/config.ts
billycao: Good question. I just copied it from `AUTH_COMMAND_TIMEOUT_MS` :P It runs asynchronously in the spawned shell and doesn't block the interface so might be okay for it to be long. You might imagine a user wants a slightly longer task and would be frustrating if the command didn't work due to a non-configurable internal timeout. But I went ahead and shortened it to 10s :) In the future if people actually use this feature can just make it configurable.

source/config.ts
billycao: Great catch, thanks!

billycao (commented):
Sorry for delay, and ty for review! :) Rebased on main and adressed comments and will merge but open to any additional feedback!

README.md
billycao: Forgot to comment earlier but done! :)

PR #175, opened 2026-03-18
feat: Add parallel tool execution support

Zebastjan:
## Summary

This PR adds support for parallel tool execution, allowing the agent to execute multiple independent tools concurrently instead of sequentially. This significantly improves performance and reduces latency when multiple operations can be performed at the same time.

### Parallel Tool Call Processing

- ✅ Support for `toolCalls` array (plural) instead of singular `toolCall`
- ✅ Index-based tool call tracking during streaming
- ✅ Proper argument accumulation across streaming deltas
- ✅ Concurrent execution of multiple independent tools

### Example Usage

**Before (sequential):**
```
User: "read package.json"
Agent: [reads file - 500ms]
User: "now read tsconfig.json"  
Agent: [reads file - 500ms]
Total: 1000ms
```

**After (parallel):**
```
User: "read package.json and read tsconfig.json and read README.md"
Agent: [reads all 3 files concurrently - 500ms]
Total: 500ms
```

### 1. Index-based streaming accumulation (`standard.ts`)

In OpenAI's streaming format, tool calls arrive in deltas:
- **First delta**: `{ index: 0, id: "call_123", function: { name: "read", arguments: "" } }`
- **Subsequent deltas**: `{ index: 0, function: { arguments: '{"file' } }` (no ID, just index)

The implementation:
- Uses `index` field to track tool calls during streaming
- First delta (with ID) creates the tool call entry
- Subsequent deltas (without IDs) use index to find and append arguments
- After streaming completes, validates using the actual tool ID

### 2. Plural toolCalls array in IR (`convert-history-ir.ts`)

- Changed from singular `toolCall` to `toolCalls` array
- Accumulates multiple tool calls from history items
- Enables proper representation of parallel execution in the IR

### Unit Tests
- ✅ `standard-tool-parsing.test.ts` - 6 tests for argument accumulation
- ✅ All existing test suites passing (129 tests)

### Manual Testing with Real Model (Kimi K2.5)

**Test 1: Single tool call**
```
list cwd
```
- ✅ Tool executed successfully with correct arguments

**Test 2: Multiple parallel tools (mixed types)**
```
read OCTO.md and list the current directory and run git status
```
- ✅ 3 tools executed in parallel
- ✅ Each tool received correct arguments
- ✅ All tools completed successfully

**Test 3: Stress test (10 parallel operations)**
```
read all the TypeScript files in source/compilers
```
- ✅ 10 files read in parallel
- ✅ All arguments accumulated correctly
- ✅ Results properly aggregated

[...]

PR #184, opened 2026-04-27
Parallel tool calls

reissbaker:
#175 was a little large as a vibecoded PR to review and parallel tool calling in general is a pretty tricky feature, since so much silently assumes single-tool-call-only in the Octo codebase. But, parallel tool calling is indeed a very useful feature: it's more efficient in terms of tokens, and also can just help LLMs be more accurate in general since if there's a complex edit in several different parts of the same large file, the LLM can send all the edits down in a single response rather than needing to track state across multiple turns.

This PR does the hard work of adding support for parallel tool calls. It's pretty invasive and large, unfortunately, as I uncovered many places in our codebase that assumed single-tool-calls.

There are a few things worth calling out in this PR:

1. `toolCall: ToolCallRequest` -> `toolCalls: ToolCallRequest[]` in the obvious places. There were also a couple history items / IRs where we were storing denormalized tool call data rather than full `ToolCallRequest` objects, and for simplification's sake in the compilers and history transforms I swapped those out to just store the original `ToolCallRequest`: previously they'd do single-item lookback to find the originating tool call, but that single-item-lookback isn't possible when there may be multiple tool calls, so rather than denorm+lookback-N-times I just store the entire originating tool call in the respective item/IR and don't need to do any lookback in those cases anymore.
2. We add a new `tool-skip` IR and HistoryItem. When a tool call in a batch fails validation, generally we should just retry right away... But, the LLM needs to know two things: a) which tool call failed, and b) that all other tool calls were skipped and didn't run. In theory we could edit the IR to remove the non-failed tool calls, but a) that could impact cache hits, and also b) that could poison the LLM with a history of making nonsense changes, i.e. if we only show it a single tool call when it intended to call a large batch, it will be confused since its in-domain action was to attempt a larger edit / set of edits / etc. Adding a `tool-skip` IR lets us track which tools were skipped, and provide correct markers in the compilers to show to the LLM that those tools didn't run using the various LLM API hooks for this sort of thing.
3. We also silently insert `tool-skip` IRs for any tool call that came *after* a rejected tool call. Generally rejection UX should only require a single rejection from the user, and so as soon as they reject one tool call, all subsequent ones are also rejected. Previous ones don't have to be rejected, since presumably they already ran i.e. previous edits.
4. Surprisingly, this required fairly invasive tool call and UI changes especially around diff handling! Previously, since tool calls happened only once, and were immediately turned into StaticItems, we could do things like read the underlying file and use it as the source of truth for rendering diffs *at UI layout time*. This doesn't work with parallel tool calls, since the file may be changing in between tool calls in the same batch, and by the time the tool runs and is turned into a `tool-output` for the static item renderer, the underlying file may have already been modified and the search string may no longer exist: both codepaths are racing each other, and the diff edit renderer sometimes would fail to render as a result, since it tries to find the search string in the underlying file as part of rendering updated line numbers. In order to handle this race condition, we need to store the file contents *at the time of the tool call parsing,* before any edits were made. The simplest/most elegant way I could think of doing this  is to actually store it on the parsed output from the tool call (i.e. just like we previously could read `edit.filePath`, we might want to be able to read `edit.originalFileContents`). This meant adding a `parse()` function to tool definitions, so that we could transform the LLM responses into richer data stored internally, including for example file contents at the time of parsing. Most tools just use a new `autoparse` function that is a no-op identity function, but the `edit` and `rewrite` tools have now have parsing logic to read and store the original file contents so that they can later be used for diff rendering. The mildly nice side effect of this change is that we don't use `readFileSync` during React layout anymore.

Playing around with it, it seems to work reasonably well! LLMs can make multiple edits in a single batch; people are presented with the edits one-by-one as if multiple requests were being made (aka the UX is the same); and rejection UX is identical and yet LLMs in my testing were able to understand precisely which tool calls ran, which were rejected, and which were skipped.

### DIFF
--- a/source/agent/trajectory-arc.ts
+++ b/source/agent/trajectory-arc.ts
@@ -256,6 +256,15 @@ export async function trajectoryArc({
   for (const toolCall of wellformedToolCalls) {
     try {
       await validateTool(abortSignal, transport, tools, toolCall.call, config);
+
+      // If we got this far, the tool validated successfully. Proactively push a tool-skip IR for
+      // it, in case other tool calls fail to validate (since all tool calls will be skipped if any
+      // are invalid).
+      retryIrs.push({
+        role: "tool-skip",
+        toolCall: toolCall,
+        reason: "One of your other tool calls was invalid, so no tool calls were run",
+      });
     } catch (e) {
       if (e instanceof FileOutdatedError) {
         const errorIr: TrajectoryOutputIR = await tryTransformFileOutdatedError(
@@ -317,6 +326,14 @@ export async function trajectoryArc({
               ...fn.arguments,
               ...fix,
             };
+
+            // Push a tool skip proactively
+            retryIrs.push({
+              role: "tool-skip",
+              toolCall: toolCall,
+              reason: "One of your other tool calls was invalid, so no tool calls were run",
+            });
+
             continue;
           }
         } catch {}
@@ -335,7 +352,14 @@ export async function trajectoryArc({
   }
 
   // If you have any IRs that need to be retried, retry them
-  if (retryIrs.length > 0) {
+  let needsRetry = false;
+  for (const ir of retryIrs) {
+    if (ir.role !== "tool-skip") {
+      needsRetry = true;
+      break;
+    }
+  }
+  if (needsRetry) {
     const fullRetryTrajectory = [...irs, ...retryIrs];
     handler.retryTool({ irs: fullRetryTrajectory });
     const retried = await trajectoryArc({
@@ -349,7 +373,7 @@ export async function trajectoryArc({
     });
     return {
       type: "finish",
-      irs: [...irs, ...retried.irs],
+      irs: [...fullRetryTrajectory, ...retried.irs],
       reason: retried.reason,
     };
   }
--- a/source/agent/trajectory-arc.ts
+++ b/source/agent/trajectory-arc.ts
@@ -67,7 +67,7 @@ type Finish = {
       }
     | {
         type: "request-tool";
-        toolCall: ToolCallRequest;
+        toolCalls: ToolCallRequest[];
       }
     | {
         type: "request-error";
@@ -110,6 +110,7 @@ export async function trajectoryArc({
     apiKey,
     model,
     abortSignal,
+    transport,
     autofixJson,
     messages: messagesCopy,
     handler: {
@@ -133,6 +134,7 @@ export async function trajectoryArc({
     model,
     autofixJson,
     abortSignal,
+    transport,
     tools,
     messages: messagesCopy,
     handlers: {
@@ -224,9 +226,9 @@ export async function trajectoryArc({
     };
   }
 
-  const { toolCall } = lastIr;
+  const { toolCalls } = lastIr;
 
-  if (toolCall == null) {
+  if (toolCalls == null) {
     return {
       type: "finish",
       reason: {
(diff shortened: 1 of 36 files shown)


source/app.tsx
```diff
@@ -649,12 +646,58 @@ const ToolRequestItem = React.memo(
...
+
+  return (
+    <Box flexDirection="column">
+      <ToolMessageRenderer item={currentToolReq} />
+      <ToolRequestRenderer
+        toolReq={currentToolReq}
+        config={config}
+        transport={transport}
+        onDone={() => {
+          setCurrentIndex(i => i + 1);
```
sdjidjev: this might need to be put in a `useCallback` to avoid making `runTool` run twice in 
```
  useEffect(() => {                                                                                                                                                  
    if (noConfirmationNeeded) {
      runTool({ toolReq, config, transport }).then(onDone);                                                                                                          
    }                                                      
  }, [toolReq, noConfirmationNeeded, config, transport, onDone]);      
```
if `BottomBarContent` makes `ToolRequestsRenderer` re-render.

source/agent/trajectory-arc.ts
```diff
@@ -201,11 +209,35 @@ export async function trajectoryArc({
...
+    // Insert tool skips for all of the non-malformed tool call IRs, and ensure the original order
+    // is kept in terms of input ordering vs output message ordering
+    irs = [...irs];
+    const malformed = new Map<string, ToolMalformedMessage>();
+    for (const ir of result.output) {
+      if (ir.role === "tool-malformed") {
+        malformed.set(ir.toolCallId, ir);
+      }
+    }
+    const wellformed = result.output.filter(ir => ir.role === "assistant");
```
sdjidjev: i love this name lol, but maybe would be more clear if we named this `assistantMessages`, esp since not all the toolCalls may be wellformed/malformed within it

source/agent/trajectory-arc.ts
```diff
@@ -236,134 +269,171 @@ export async function trajectoryArc({
...
-    };
-  } catch (e) {
-    if (e instanceof FileOutdatedError) {
-      const errorIrs: TrajectoryOutputIR = await tryTransformFileOutdatedError(
-        abortSignal,
-        transport,
-        toolCall,
-        e,
+  let retryIrs: TrajectoryOutputIR[] = [];
+  const wellformedToolCalls: Array<ToolCallRequest> = [];
```
sdjidjev, line 258: even though i love the wordplay involved with `wellformed`, maybe we could also name it `validToolCalls` instead to match `validateTool`, etc.

sdjidjev (commented):
didn't do a 100% full pass yet but so far looking great!! excited for parallel tool calls!!

will do the remainder of the review later today, just wanted to get my initial skim review out there!

source/app.tsx
reissbaker: Oooo good catch

source/agent/trajectory-arc.ts
```diff
@@ -236,134 +269,171 @@ export async function trajectoryArc({
...
+            fn.arguments = {
+              ...fn.arguments,
+              ...fix,
+            };
+
+            // Push a tool skip proactively
+            retryIrs.push({
+              role: "tool-skip",
+              toolCall: toolCall,
+              reason: "One of your other tool calls was invalid, so no tool calls were run",
```
sdjidjev: Based on your PR description shouldn't the phrasing be more like:
"One of your past tool calls was invalid, so all tool calls afterwards (including this one) were not run."
since we wouldn't have skipped a tool call before the invalid one?

sdjidjev: Also to dedupe since it's used in multiple places this string could be in `ir-prompts.ts`

source/agent/trajectory-arc.ts
reissbaker: Good point! I simplified the IR and compiler interface overall based on a TODO comment I left, which means Compiler interfaces only return AssistantMessages now, since AssistantMessage toolCalls can include MalformedRequests. This means we don't need to do any filtering on the return type of the compiler, since it's always an AssistantMessage, so this variable (and entire code block) has been deleted.

source/result.ts
```diff
@@ -0,0 +1,9 @@
+export type Result<T, E> =
+  | {
+      success: true;
+      data: T;
+    }
+  | {
+      success: false;
```
sdjidjev, line 7: what would be a scenario where `success` would be `false` for parsing? just wondering if we could simplify this.

source/agent/trajectory-arc.ts
reissbaker: So, the tricky thing here is that these *aren't* validated tool calls: they haven't had `validateTool(...)` called on them yet, so I think calling them `validToolCalls` is a little more confusing (why do we then need to re-validate the already-valid tool calls?). At this point in the code all we know is that they're not malformed (i.e. garbled output). I'm down for a different name if you still think it's worth one, open to suggestions on it.

source/app.tsx
reissbaker: Fixed + made canary builds throw a hard error if `runTool` is called when a tool is already running.

source/tools/tool-defs/list.ts
```diff
@@ -8,7 +8,7 @@ const ArgumentsSchema = t.subtype({
 const Schema = t
   .subtype({
     name: t.value("list"),
-    arguments: t.optional(ArgumentsSchema),
```
sdjidjev, line 11: hmm wonder why was this optional in the first place 🤔

source/agent/trajectory-arc.ts
sdjidjev: Ohh I see... 🤔 
I'm unsure then on what should happen in the case of a malformed tool call vs. a invalid tool call. When do we execute `tool-skip`s?

EDIT: your explanation above helped me understand, thanks!

sdjidjev (approved):
Code looks good to me!! Just had a few comments/questions I left.

Also, the `parse` solution you came up does seem to be the most elegant way to handle the UI rendering issue but I do wonder if there could be a more obvious way to handle it.
Like if someone were to add a tool that doesn't behave politely in a parallel tool call situation, how could we make it obvious that they need to add a custom `parse` fn in that tool?

I also haven't tried in detail the parallel tool calls yet but gonna do that now, excited!

source/agent/trajectory-arc.ts
reissbaker: So this codepath is just basically `assert no malformed requests exist anymore` since we handled them above, but the type system doesn't know that and so we're copying everything over into a narrowly-typed array!

For `tool-skip` IRs, we potentially generate them in two places:

1. Above, where we handle malformed tool requests. All well-formed tool requests get `tool-skip` IRs associated with them if any tool requests are malformed.
2. Below, where we handle tool validation. All valid tool requests get `tool-skip` IRs associated with them if any tool requests fail validation.

source/agent/trajectory-arc.ts
reissbaker: Ah great point on the dedupe! But no — only *rejections* cause subsequent tool calls to be skipped while preserving previous tool calls. During validation, *all* tool calls that pass validation (well-formedness, or validation) will be skipped if any of them fail. From a UX standpoint, it's a little weird to be presented with a half-valid batch of requests: we might as well just retry the entire batch, and let the LLM know specifically which ones failed and which ones were fine but skipped due to failure.

Basically:

## Rejection flow:

- LLM makes a batch of tool call requests a, b, c
- User accepts a
- We run a  and create a tool-output IR associated with it for its output
- User rejects b with message X
- We create a rejection IR with message X associated with b, and a tool-skip IR for c which never ran
- We call the LLM to generate the next response

In this flow, the `tool-skip` IRs only are subsequent to rejection IRs, since we already executed previous tool call requests prior to getting a rejection.

## Validation flow

- LLM makes a batch of tool call requests a, b, c
- We validate each of them: a passes, b fails, c passes
- We create the following IRs: a tool-skip for a, an invalid tool for b, and a tool-skip for c
- We automatically call the LLM to retry

In this flow, the `tool-skip` IRs aren't necessarily subsequent to the failed ones; they can be in any order.

sdjidjev: Ohhhh I see, thank you for explaining.

I realize now I was conflating the words `rejected` and `failed` in the PR summary which was getting me confused 😅 

Separating the flows like this helped, thanks!!

sdjidjev: If I'm understanding correctly, would we need to make a different `reason` for the `tool-skip` in the case of your **Rejection flow:**

reissbaker: Yup, exactly! In `state.ts` in the rejection flow, we set a different reason :) https://github.com/synthetic-lab/octofriend/pull/184/changes#diff-12bf53877f009f752ce4b4d9f482fdd8a71df3950c6f29a8dfa25c0b8910180eR221

source/result.ts
reissbaker: Hmmm I could imagine reasons we might return `success: false` for parsing although it's true the current code doesn't do that (since `validate` already checks the file exists). In theory we probably should still catch file read errors and return `success: false` though just in case of race conditions e.g. two agents running at the same time, and one deletes the file right after the other validates its existence but before it actually reads the file. I'll add that.

In general I want to move more towards Rust-like `Result`s rather than throwing errors for most cases, since crashes are bad and Results force you to do error handling whereas you may not realize if a codepath can throw an error / tsc can't help you.

sdjidjev (approved):
Just tried it out in various scenarios, it's so good 🎉 🎉

source/tools/tool-defs/list.ts
reissbaker: Basically to handle dumb LLMs that couldn't figure out to send an empty hash for arguments. AFAICT all the major ones we host are fine now, and it complicates the types to allow for optional arguments so I removed it

source/result.ts
reissbaker: Done! We now handle file read errors and return `success: false` for them

PR #185, opened 2026-04-28
Fix shell timeout with bash commands

sdjidjev:
Previously when running a command such as `npm run dev &`, the command hangs forever and doesn't time out (even though Octo does set a timeout).

The fix is to manually run a `setTimeout` and kill the process group once `timeout` is hit.
The built-in `timeout` option in `spawn` won't work since it would just kill the child process `bash` instead of the whole group (including `npm` in this case).

### DIFF
--- a/source/transports/local.ts
+++ b/source/transports/local.ts
@@ -80,16 +80,15 @@ export class LocalTransport implements Transport {
       const child = spawn(cmd, {
         cwd: process.cwd(),
         shell: "bash",
-        timeout,
         stdio: ["ignore", "pipe", "pipe"],
         detached: true,
       });
 
       let output = "";
       let aborted = false;
+      let timedOut = false;
 
-      const onAbort = () => {
-        aborted = true;
+      function killGroup() {
         // Kill the entire process group to handle child processes
         try {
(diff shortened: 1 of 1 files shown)


source/transports/local.ts
```diff
@@ -80,16 +80,15 @@ export class LocalTransport implements Transport {
...
         detached: true,
       });
 
       let output = "";
       let aborted = false;
+      let timedOut = false;
 
-      const onAbort = () => {
-        aborted = true;
+      function killGroup() {
```
reissbaker, line 92: Since this gets called potentially multiple times (both in the timer and `onAbort`), probably worth a:

```typescript
let killed = false;

function killGroup() {
  if(killed) return;
  killed = true;
  // ...
}
```

*Usually* this won't matter, but if someone spawned a group after the abort that happened to get the now-free group ID, technically you might accidentally kill some other random process group on the machine as a result.

reissbaker (approved):
Makes sense + LGTM! Left one minor nit.

source/transports/local.ts
sdjidjev: Ooh good catch, fixed!

PR #186, opened 2026-04-29
Fix package.json resolution in unit tests

billycao:
`autocompact.test.test` fails with:

CleanShot 2026-04-28 at 17 41 23@2x

`path.join(__dir, "../../package.json")` works when run from `dist/` but not from source. This just steps up the tree twice to look for package.json.

### DIFF
--- a/source/config.ts
+++ b/source/config.ts
@@ -616,10 +616,15 @@ export type Metadata = {
 };
 
 async function readMetadata(): Promise<Metadata> {
-  const packageFile = await fs.readFile(path.join(__dir, "../../package.json"), "utf8");
-  const packageJson = JSON.parse(packageFile);
+  const candidates = [path.join(__dir, "../package.json"), path.join(__dir, "../../package.json")];
 
-  return {
-    version: packageJson["version"],
-  };
+  for (const candidate of candidates) {
+    if (await fileExists(candidate)) {
+      const packageFile = await fs.readFile(candidate, "utf8");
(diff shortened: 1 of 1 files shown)


billycao:
Merging to test a github actions workflow for unit tests but open to feedback!

sdjidjev:
@billycao 
```
vi.mock("../config.ts", () => {
  return {
    APP_METADATA: {
      version: "0.0.42",
    },
  };
});
```

I think it's probably better to put this in the broken test than to change the `config.ts`
Haha barely missed the merge 😅

billycao:
> @billycao
> 
> ```
> vi.mock("../config.ts", () => {
>   return {
>     APP_METADATA: {
>       version: "0.0.42",
>     },
>   };
> });
> ```
> 
> I think it's probably better to put this in the broken test than to change the `config.ts` Haha barely missed the merge 😅

Good point! I get not affecting prod code for tests but I'd argue that if in the future `Metadata `changes, manually defining it for tests may require you to keep it up to date, or else the test might not testing what's actually in production. 

For now I just added it as

```
vi.mock("../config.ts", () => {
  const APP_METADATA: Metadata = {
    version: "0.0.42",
  };

  return { APP_METADATA };
});
```

to encourage keeping it updated. Reverted the original PR commit and added new code in https://github.com/synthetic-lab/octofriend/commit/cc8af34e9bf9e77bedd0a48a885cf3625a4adea9

PR #187, opened 2026-04-29
Github actions workflow to run unit tests on runs-on

billycao:
Runs on PRs and commits to `main`. Uses our existing runs-on (AWS spot instance) runner infra


reissbaker (approved):
Sweet! FYI, I changed the settings for running workflows to require our permission to run, so we don't get accidentally DOSed:

Image

PR #188, opened 2026-04-30
Improve notifications

reissbaker:
#183 was pretty nice, but I think it's a little confusing that we don't send notifications when Octo is waiting for things like tool prompts/permissions. Also, it's a little spammy, since it notifies *every* time Octo is done responding, even if you're actively interacting with it.

This PR improves notifications for Octo by:

1. Notifying for any state where Octo needs input from you, including tool prompts
2. Waiting a configurable amount of time before sending notifications, and canceling the notifications on input. By default, the wait time is 10s. This feels like a reasonable default in terms of not wasting *too* much time waiting, while also avoiding most notification spam.

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,22 +259,36 @@ OCTO_VERBOSE=1 octofriend
 
 ## Notify Finish Command
 
-You can configure a command to run whenever Octo finishes responding to a prompt.
-This is useful for desktop notifications, sounds, or other integrations.
+You can configure a command to run whenever Octo finishes responding to a
+prompt or needs input like permissions. This is useful for desktop
+notifications, sounds, or other integrations.
 
-This command will **not** run for tool call confirmations. It will only run when
-Octo is ready for a new prompt (eg. after responding to a prompt in unchained mode)
-
-Add a `notifyFinishCommand` to your `~/.config/octofriend/octofriend.json5`:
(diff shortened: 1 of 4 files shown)


source/config.ts
sdjidjev: we should bump this value up!

README.md
```diff
@@ -257,24 +257,38 @@ underlying error messages from APIs or tool calls, run Octo with the
...
+},
+```
+
+By default, Octo will wait 10 seconds before notifying you, and if it receives
+input during that time it'll skip the notification (so as to not spam you with
+notifications when you're actively attending to it and chatting). To change the
+wait time, set:
+
+```json5
+notification: {
```
sdjidjev: -typo since ppl might copy/paste- `notifications`

source/config.ts
sdjidjev: need to update this to use the new schema!

reissbaker: oops! will do!

reissbaker: whoops sloppy i forgot to delete the old thing, deleted it which made tsc complain about this line and fixed this line

README.md
reissbaker: fixed!

sdjidjev (approved):
sw8!

PR #190, opened 2026-05-01
Add grep tool

reissbaker:
Adds a `grep` tool similar to our `glob` tool, based on native POSIX grep. Closes https://github.com/synthetic-lab/octofriend/issues/138

Useful so that Octo doesn't prompt for perms when it's just trying to do safe filesystem searching.


sdjidjev (approved):
Yay so much better than shelling out each time 🐚

PR #194, opened 2026-05-04
Fix tool-failed UI message

reissbaker:
#184 accidentally introduced a UI bug for tool call failures: instead of saying the tool call failed, it says the tool call *validation* failed. For example, if Octo tries to run `npx tsc`, and it runs, but `tsc` notices a type error in the codebase, instead of saying the tool failed, instead it will say that tool call validation failed. But it didn't fail validation: it validated fine, and even ran the command. The command just exited with a non-zero status code, since there's a type error.

This is because we reuse the same `tool-error` IR for both validation failures, and for failures when the tool call itself has an error, e.g. a shell command exits with a non-zero status code. The UI can't tell the difference, since the IR is overloaded and could mean either scenario.

This PR adds a new IR to distinguish between validation errors and actual underlying tool run failures. Here's what it now renders as:

image

Arguably I could've reused the malformed tool IRs. However, the nice thing that #184 introduced with those is better type safety: they reference the original MalformedRequest from the assistant message. But for tool call *validation* failures, there is no MalformedRequest! The request was well-formed, but some validation code decided it was no good anyway. Adding the new IR type helps preserve the type safety niceties there while still allowing us to distinguish in the UI between different failure points for tool calls.

### DIFF
--- a/source/app.tsx
+++ b/source/app.tsx
@@ -980,7 +980,7 @@ const MessageDisplayInner = React.memo(({ item }: { item: HistoryItem | Inflight
           <Text color="red">
             {displayLog({
               verbose: `Error: ${item.error}`,
-              info: "Tool validation failed...",
+              info: "Tool failed...",
             })}
           </Text>
         </Box>
--- a/source/app.tsx
+++ b/source/app.tsx
@@ -973,6 +973,18 @@ const MessageDisplayInner = React.memo(({ item }: { item: HistoryItem | Inflight
       </Text>
     );
   }
+
+  if (item.type === "tool-validation-error") {
+    return (
+      <Text color="red">
+        {displayLog({
+          verbose: `Error: ${item.error}`,
+          info: "Tool call failed validation checks. Retrying...",
+        })}
+      </Text>
+    );
+  }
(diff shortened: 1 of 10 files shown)


source/app.tsx
```diff
@@ -973,14 +973,26 @@ const MessageDisplayInner = React.memo(({ item }: { item: HistoryItem | Inflight
       </Text>
     );
   }
+
+  if (item.type === "tool-validation-error") {
+    return (
+      <Text color="red">
+        {displayLog({
+          verbose: `Error: ${item.error}`,
+          info: "Tool call failed validation checks. Retrying...",
```
sdjidjev: -minor nit- but since tool-validation-error could occur as part of an abort, it would make it so the`Retrying...` message wouldn't be applicable in that case

maybe an `aborted` field might be useful for discernment?

sdjidjev (approved):
LGTM just one nit

source/app.tsx
reissbaker: Great point

reissbaker: done!

PR #195, opened 2026-05-04
Fix multi-line input Static Rendering

sdjidjev:
### Before fix:
| Input | Static Content |
|----------|----------|
| Screenshot 2026-05-04 at 3 56 10 PM | Screenshot 2026-05-04 at 3 56 15 PM |

### After fix:
| Input | Static Content |
|----------|----------|
| Screenshot 2026-05-04 at 4 26 49 PM | Screenshot 2026-05-04 at 4 26 54 PM |

Pasting content with several newlines in them would lead to weird character removal behavior and spacing issues. By splitting on the same regex we use for the input, the Static Content afterwards looks good!


reissbaker (approved):
Looks great!

PR #197, opened 2026-05-07
Add line numbers to file read tool

sdjidjev:
For more accurate reads, adds line numbers at the beginning of each line as part of the `content` string for the `read` tool.


reissbaker (approved):
Hell yeah!

PR #198, opened 2026-05-07
Fix InputWithHistory multi up-arrow bug

sdjidjev:
With [react-ink](https://github.com/vadimdemedes/ink/)'s 7.0.0 upgrade, our history feature (pressing up/down arrows to access input history) was broken past the first up-arrow press.

By accessing the currentIndex and originalInput values with `useRef` instead of `useState`, we can guarantee a fresh read on the object within our handlers.


reissbaker:
Closing in favor of #199 which fixes the underlying problem! Turns out this was a React bug with `useEffectEvent` that would have messed up every `useInput` subtly...

PR #201, opened 2026-05-13
Flattens glob/grep arguments

sdjidjev:
This format plays more nicely with GLM. Nested objects makes it return malformed JSON right now.

Quick patch to allow these tools to work in octo!


reissbaker (approved):
LGTM!

sdjidjev:
Screenshot 2026-05-12 at 5 15 33 PM
grep + glob

PR #203, opened 2026-05-20
feat: support partial file reads

puneetdixit200:
Closes #154.

## Summary
- Adds optional `offset` and `limit` arguments to the `read` tool for line-range text reads.
- Keeps partial reads untracked for edit safety, so an edit still requires a current full-file read.
- Preserves full-read behavior for images and for text files that have already been fully read, and keys partial read context by file plus range to avoid collapsing different ranges.
- Adds regression tests for ranged output, edit-safety behavior, and full-read upgrade behavior.

## Verification
- `npx vitest run source/tools/tool-defs/read.test.ts`
- `npm run build`
- `npx prettier --check source/tools/tool-defs/read.ts source/tools/tool-defs/read.test.ts source/ir/convert-history-ir.ts`
- `npm run test:run`

AI assistance was used to prepare this change. I reviewed the implementation against the existing file-tracker safety model and verified it with the commands above.


reissbaker (changes requested):
This is a very useful feature, but I think the current implementation falls prey to the problem described here: https://github.com/synthetic-lab/octofriend/issues/154

Specifically, because we filter out "stale" file reads in the compilers for prompt optimization purposes, this implementation could end up causing partial-file-reads to filter out earlier full-file-reads, and repeated partial-file-reads would also filter each other out, since the compilers aren't aware of the existence of partial file reads.

We have a pretty large refactor of the LLM IR and the tools in flight in #206 — I think once that's merged, it'll be much simpler to do this. In #206, the tools themselves decide whether to return `file-read` IRs vs just returning ordinary content; since only the `file-read` IRs are subject to the prompt optimization / filtering, we could make the `read` tool only return those for full-file reads, and instead return ordinary tool content for partial file reads (which would then be immune from filtering).

reissbaker:
FYI now that #206 is merged, I'd recommend just porting this on top of it :) You won't have to mess around in the history-IR conversion and can just return ordinary tool content from `read` for the partial file reads, instead of returning the custom IR!

puneetdixit200:
Thanks, this is now ported on top of the merged tool/IR changes. Partial reads return ordinary tool output via the untracked read path, while full reads still return file-read IR for edit safety.

I re-ran:
- `npx vitest run source/tools/tool-defs/read.test.ts`
- `npx prettier --check source/tools/tool-defs/read.ts source/tools/tool-defs/read.test.ts`
- `npm run build`
- `npm run test:run`

reissbaker (approved):
Beautiful! Thanks so much @puneetdixit200 :)

PR #206, opened 2026-05-21
Port Octo to generic IR

reissbaker:
This PR is the final-ish step of the changes started in #200 : it ports Octo to actually use the generic IR and tool definitions in `libocto`, and removes the old IR. It also removes most of the old `history.ts`, which sucked due to it reimplementing the original LlmIR mostly-pointlessly, and replaces almost all HistoryItems with a simple:

```typescript
{
  type: "llm-ir",
  ir: OctoIR,
}
```

The last few HistoryItems that exist are UI state that are unrelated to data we'd pass to LLMs, and so there's no duplicated code anymore between the two.

This makes Octo basically structured very similarly to a compiler: there are high-level, Octo-specific intermediate representations (IR) of agentic behavior called `OctoIR`, which are compiled down to lower-level generic representations called `LoweredIR`, and finally those are compiled down to some "native" API (the standard OpenAI Chat Completions API, the newer OpenAI Responses API, or the Anthropic API). The native compiler stage is basically a set of very simple transforms from `LoweredIR` to the native forms, thanks to the compiler simplification in #202 which is largely preserved in this port.

Actually using the generic IR from #200 required a few changes to `libocto` since there were a couple ergonomic issues around the extension types; however general usage is largely similar to the initial impl in #200. The biggest change was to the tool builder DSL; to make it easier to define tools that operate on custom IR without needing circular type imports, the DSL changed to separate the builder step into three:

1. A `declare` function that first defines the schema,
2. An optional `withCustomIR` on a set of named functions that define custom IR, so that later calls can just use those named functions, and so that the custom IRs themselves don't need to import the tools;
3. A `define` function that implements the tool validation/parsing/running functions, so that those functions don't need to circularly depend on the types.

Generally speaking this should be pretty easy to see via example in the tool definitions in `source/tools/`. Here's the `read` tool, for example, which uses a custom file read IR:

```typescript
import { t } from "structural";
import { fileTracker } from "../file-tracker.ts";
import { attempt, attemptUntrackedStat, TOOL, fileReadIR } from "../common.ts";
import { isImagePath, loadImageFromPath } from "../../utils/image-utils.ts";

function withLineNumbers(content: string): string {
  return content
    .split("\n")
    .map((line, i) => `${i + 1}: ${line}`)
    .join("\n");
}

const read = TOOL.declare({
  name: "read",
  description: `
Reads file contents as UTF-8, or loads supported image files (PNG, JPEG, etc.) for visual
inspection. Prefer this to Unix tools like \`cat\`. Text output is prefixed with line numbers in
the form \`N: content\` so you can refer to exact positions; the line-number prefix is NOT part
of the file and must not be included when constructing edit/search strings.
`.trim(),
  ArgumentsSchema: t.subtype({
    filePath: t.str.comment("Path to file to read"),
  }),
});

export default read.withCustomIR({ fileReadIR }).define(async () => ({
  async validate(abortSignal, transport, toolCall) {
    return await attemptUntrackedStat(transport, abortSignal, toolCall.parsed.arguments.filePath);
  },
  async run({ signal, transport, toolCall, customIR }) {
    const { filePath } = toolCall.parsed.arguments;

    if (isImagePath(filePath)) {
      return attempt(`Could not read image ${filePath}`, async () => {
        const resolvedPath = await transport.resolvePath(signal, filePath);
        const image = await loadImageFromPath(resolvedPath);
        await fileTracker.recordFileReadTimestamp(transport, signal, filePath);
        return customIR.fileReadIR({ content: `Image file: ${resolvedPath}`, image });
      });
    }

    return attempt(`No such file ${filePath}`, async () => {
      const content = await fileTracker.read(transport, signal, filePath);
      return customIR.fileReadIR({ content: withLineNumbers(content) });
    });
  },
}));

```

This also fixes a :warning: major bug :warning: that I discovered in Octo, which has existed since the beginning of time... Previously, any `.comment(...)` calls on the top-level Schema were discarded, and only the comments on the ArgumentsSchema (or child fields) made it through to the LLM! This bug impacted:

1. bash
4. create
5. fetch
6. glob
7. grep
8. list
9. LSP tools
10. MCP
11. read
12. skill
13. web-search

Whoops. Apparently the underlying models still knew enough or had enough context to figure out how to use the tools anyway, but uhh they should work a lot better in this PR now that we're not dropping the majority of tool descriptions...

ANYWAY. Octo uses `libocto`'s IR extensions to define a few extra IRs that aren't in `libocto`:

1. `file-mutate`
14. `file-read`
15. `tool-reject`

These are just simple ports of Octo's original IRs with these names. They're not in libocto currently since libocto only defines a base IR and a DSL for defining agents and tools; it doesn't define any tools themselves, and these all relate to specific tools/permissions that different clients may not use or have.

In the long run, we'd likely move Octo's tool definitions into reusable, optional libraries in `libocto`, which means `file-mutate` and `file-read` would also move to optional extensions defined in `libocto` (automatically inherited if you use any of the file tools). `tool-reject` feels less likely to move into `libocto` core, since not all clients have the concept of tool permissioning or rejections; for example, the Synthetic web UI wouldn't likely expose permissioning, and the goal of `libocto` is to be relatively minimal in terms of the IR AST and only require you to handle IR that's actually relevant to what you're building. Octo needs the concept of "rejecting a tool," so it would continue to define it as an IR extension for itself, but other clients (like Synthetic) wouldn't.

We have a couple lowering functions that take higher-level IR and lower them into base IR that the compilers understand:

1. `optimizeFiles`, which lowers the file-based IRs and automatically applies our prompt optimizations.
4. `lowerTrajectories`, which currently does nothing, since subagent trajectories aren't implemented yet in Octo.
5. `lowerToolRejects` which converts Octo's custom `tool-reject` IRs into the underlying base IRs (`tool-skip-output`) with an appropriate message.

The compilers and autocompaction have all been ported to operate on the low-level IRs from `libocto`, and so you're forced by `tsc` to call these lowering functions before passing any of the high-level `OctoIR` to the lower-level compilers.

Since tools all now return `Result`s rather than throwing errors, trajectories were cleaned up to stop catching most errors and instead inspect result return types. I noticed that the:

```typescript
export const result = {
  ok(...) {
  },
  err(...) {
  }
};
```

Helper const was actually annoying to use in practice, since it meant you had to `import { result } from "./result.ts"`, which forces `result` to suddenly be a reserved word, and `const result = await someResultFn(...)` is really nice from a variable naming function, so I removed the old `result` export and replaced it with just exporting the `ok` and `err` functions, leaving `result` free as a variable name.

This is generally a massive PR, since it effectively rewrites Octo's core and a *lot* of machinery had to be ported (usually in minor ways, e.g. `app.tsx` needing to use OctoIR instead of the duplicated HistoryItem definitions). Sorry for the difficult review! I think this should unblock working on most other big features (like session resumption) since we shouldn't need to do a massive rewrite again.

I'm at least fairly pleased that despite adding a bunch of tests, this PR is still an overall reduction in lines of code, since it removes so much duplication.

## Future work

This is pretty much all that needs to be done *for Octo*, but `libocto` itself isn't quite ready for easy consumption. The main missing pieces:

1. The compilers that turn `LoweredIR` into native API calls aren't in `libocto`: they're in Octo itself. Currently the `Compiler` type depends on `ModelConfig` to get the model details for the native API calls, which forces a dependency on Octo's config file; this should be straightforward to factor out, I just didn't want to make this PR even bigger than it already is.
2. The tools should probably move into `libocto` as optional imports you can use, so that clients don't have to reimplement e.g. file reads, web searches, etc.
3. Autocompaction is still pretty dependent on some Octo-specific configuration that we should factor out so that `libocto` has automatic compaction built in.
4. Trajectories are *very* Octo-dependent. Factoring them out is a little more annoying due to edit autofixing living in the trajectory code, and trajectories should be generic over tools (i.e. not know whether the edit tool is in use). This probably requires a minor extension of `libocto` to allow tools to define optional `autofix` functions, so that edit tool autofixing can live on the edit tool rather than in trajectories.
5. Subagents don't really exist per se. Most of the  machinery is there, but I probably need to do some more thinking on flexible ways to decide how to lower trajectories into the `LoweredIR`; what does the top-level agent see from a "finished" subagent, and how do we deterine the subagent is done? One option is that the final assistant message is the subagent's "response" to the main agent (and it isn't done until it makes an assistant message with no tool calls), although I think there are probably flexible cases where we'd want the subagent to be forced to call some tool/function in order to stop, and hardcoding that the final thing as an assistant message feels a little forced. We should probably build out various subagents in Octo in order to explore this design space.
6. `libocto` and `octofriend` should probably be sister packages in an npm monorepo/workspace, so that you don't need to depend on (for example) `ink` in order to use `libocto`. Many `libocto` clients may not be terminal applications.


source/tools/tool-defs/mcp.ts
```diff
@@ -128,120 +114,135 @@ export async function shutdownMcpClients(): Promise<void> {
...
+      const client = await withAbort(getMcpClient(serverName, data));
+      if (!client.success) return client;
 
-        // List available tools to check if the requested tool exists
-        const tools = await withAbort(client.listTools());
-        const tool = tools.tools.find(t => t.name === toolName);
-
-        if (!tool) {
-          const availableTools = tools.tools.map(t => t.name).join(", ");
-          throw new ToolError(
```
sdjidjev, line 178: i'm a little confused on when we throw an error from a tool versus returning `err` from `result.ts`, since i see both being used. in which situations would you use either?

source/compilers/anthropic.ts
```diff
@@ -172,15 +173,14 @@ function modelMessageFromIr(
...
-  if (ir.role === "compaction-checkpoint") {
+  if (ir.role === "checkpoint") {
     return {
       role: "user",
-      content: compactionCompilerExplanation(ir.summary),
+      content: anthropicContentParts(ir.content, modalities),
     };
   }
 
-  const _: never = ir;
```
sdjidjev, line 182: unless i'm missing something, wouldn't we still need this to ensure exhaustive use of the ir role? 
right now when i delete the `checkpoint` one it  compiles & runs fine.

source/libocto/tool-def.ts
```diff
@@ -317,11 +268,11 @@ export class ToolBuilder<Extra = never, Data = unknown> {
   dynamicDefineTool<
     T extends (
       args: ToolFactoryArgs<Data>,
-    ) => Promise<ToolFactory<Data, any, any, any, Extra> | null>,
+    ) => Promise<ToolFactory<Data, any, any, any, any> | null>,
```
sdjidjev: idk if this is too crazy of a suggestion but i feel like it might be nice to annotate what each of these`any`s represent. either with JSDoc stuff or like 
```
type AnyToolSchema = Schema<any, any>;
type AnyToolParsedArguments = any;
type AnyToolData = any;
type AnyToolSubagentNames = any;
type AnyToolCustomIRData = any;
```
the way you have it now, someone building out tools in the future won't need to know that info, but i feel like annotating it would make future work within `tool-def.ts` easier to understand!

source/tools/index.ts
```diff
@@ -27,8 +23,9 @@ export async function loadTools(
 
   await Promise.all(
     (Object.keys(toolMap) as Array<keyof typeof toolMap>).map(async key => {
-      const toolDef = await toolMap[key](signal, transport, config);
+      const toolDef = await toolMap[key]({ signal, transport, data: config });
       if (toolDef) {
+        (toolDef as any).name = key;
```
sdjidjev: -nit- i don't think you have to cast it as `any`!

source/tools/index.ts
```diff
@@ -75,16 +82,22 @@ export async function validateTool(
   loaded: Partial<LoadedTools>,
   tool: ToolCall,
   config: Config,
-): Promise<null> {
+): Promise<Result<null, string>> {
   const toolDef = lookup(loaded, tool);
-  return await toolDef.validate(abortSignal, transport, tool.original, config);
+  if (!toolDef.success) return toolDef;
+  return await (toolDef.data as any).validate(
```
sdjidjev: -nit- i think the `any` cast here is redundant too!

sdjidjev (approved):
Wow so good!!! Just left a few nits/questions but otherwise LGTM! 🐙

source/tools/tool-defs/mcp.ts
reissbaker: Whoops, good catch! In the new world, tools should never throw an error: they should only return `err`. I'll fix the couple spots that are still throwing errors (I think those are just re-throwing unknown errors when wrapping in a `try`/`catch`, and some of those are inside an `attempt` which autoconverts thrown errors into `err` calls, but  there are a couple missing spots and anyway the throw-and-rely-on-`attempt` is confusing).

source/compilers/anthropic.ts
reissbaker: OH good point, gpt-5.5 got a little aggro on removing these. Done

source/tools/index.ts
reissbaker: great catch, fixed

source/tools/index.ts
reissbaker: great catch, fixed

source/libocto/tool-def.ts
reissbaker: good point! done

reissbaker:
Thanks for the review @sdjidjev :) I think all the comments are addressed!

sdjidjev:
Woo looks great, ship it ⛵🐙

PR #207, opened 2026-05-27
Port compilers to libocto

reissbaker:
As per my "Future work" comments on #206 , this is now the future and I have ported the compilers to `libocto` :P They no longer directly reference Octo's config, and instead have their own config options that we hydrate at runtime in the high-level, Octo-specific `run` function from the config.

While doing this I also tore out the Vercel AI SDK used for the responses API, and replaced it with ordinary OpenAI calls to reduce dependencies and keep us closer to the metal. This required adding some OpenAI-specific JSON Schema patching, since the OpenAI API actually does *not* conform to JSON Schema when using `strict` mode, and will throw errors if you use optional keys or leave enums untyped. Who knows what the Vercel SDK was doing before, since these are not semantic-free changes... Anyway, Octo now converts to and from the OpenAI-specific JSON Schema subset and the full JSON Schema that Structural supports, and uses strict mode function calling so that it's impossible for OpenAI models to incorrectly call a tool.

In the future this should also unblock using OpenAI subscriptions inside Octo, which would require us hitting their special subscription API directly with HTTP requests and doing these transforms ourselves. OpenAI seems to be okay with people doing that (e.g. explicitly allowed OpenCode to do it), so I think that should be fine. We can look at OpenCode's auth code for examples.

The main `run(...)` function is still a high-level, Octo-specific function that does the lowering and calls the lower-level `libocto` compilers. This is probably correct? My sense is that whatever future subagent-aware trajectory code we eventually write and port into `libocto` will have some `compiler: ...` required option where you give it a high-level compiler that takes your agent's IR and compiles it down to low-level `libocto` IR.


sdjidjev (approved):
LGTM!

PR #208, opened 2026-05-28
Fix Synthetic quota display

reissbaker:
When we changed to a weekly credit limit and a rolling five-hour window for requests, we never updated Octo, and so Octo always kept saying you had used zero requests (since it kept checking the old, deprecated fields).

This fixes that:

image

### DIFF
--- a/source/utils/quota.ts
+++ b/source/utils/quota.ts
@@ -1,41 +1,69 @@
 import { t } from "structural";
 
-export type QuotaEntry = {
-  used: number;
-  limit: number;
-  renewsAt: Date;
+const QuotaEntryRawSchema = t.subtype({
+  remaining: t.num,
+  max: t.num,
+  nextTickAt: t.str,
+  tickPercent: t.num,
+});
+
+const WeeklyEntryRawSchema = t.subtype({
+  nextRegenAt: t.str,
+  percentRemaining: t.num,
+  maxCredits: t.str,
+  remainingCredits: t.str,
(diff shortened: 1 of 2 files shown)


source/utils/quota.ts
```diff
@@ -1,41 +1,69 @@
...
+  remainingCredits: t.str,
+  nextRegenCredits: t.str,
+});
+
+type WithDateFields<T extends {}, K extends keyof T> = Omit<T, K> & {
+  [P in K]: Date;
 };
 
+export type QuotaEntry = WithDateFields<t.GetType<typeof QuotaEntryRawSchema>, "nextTickAt">;
+export type WeeklyEntry = WithDateFields<t.GetType<typeof WeeklyEntryRawSchema>, "nextRegenAt">;
```
sdjidjev, line 23: i forget 😅 why did we call one `nextTickAt` and the other `nextRegenAt`?

sdjidjev (approved):
Looks great! Just had a question for my own knowledge 🌠

source/utils/quota.ts
reissbaker: I'm dumb. Doesn't seem worth breaking clients that already consume the API but, I'm dumb. We normalize it in the UI.

reissbaker: Other dumb things: note that in one case the percentage is a number from 0-1, and in another case the percentage is 0-100. Bad.

PR #211, opened 2026-06-02
Clarify glob and grep tool arguments

sdjidjev:
It seems that LLMs didn't understand the different arguments so renamed them and was more explicit about their usecases in their `comment`s.

### Glob argument changes

| New | Old | Notes |
|-----|-----|-------|
| `path` | ~`cwd`~ ||
| `includeName` | ~`name`~ ||
| `excludeName` | ~`excludeFilename`~||
| `includePath` | ~`path`~ ||
| _removed_ | ~`type`~ |All comments referenced finding files, so seemed unnecessary to surface the option|

### Grep argument changes

| New | Old | Notes |
|-----|-----|-------|
| `path` | ~`cwd`~ ||
| `pattern` || Internals & comment updated to allow extended regex with `-E` flag|

### Other changes
- Grep used to check the error string for `"exit code 1"` for empty results but the actual string was actually "`Command exited with code: 1`"
  - We now check the `exitCode` passed through in the error

### DIFF
--- a/source/app.tsx
+++ b/source/app.tsx
@@ -1224,9 +1224,9 @@ function GlobRenderer({ item }: { item: ParsedToolSchemaFrom<typeof glob> }) {
   return (
     <Box flexDirection="column">
       <Text color="gray">Octo searched for files using a glob pattern:</Text>
-      <GlobArg name="CWD" arg={item.arguments.cwd} />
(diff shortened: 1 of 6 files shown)


reissbaker (approved):
LGTM! Good fixes.

PR #212, opened 2026-06-12
Autocompaction inside libocto

reissbaker:
In my ongoing quest to pull most of the "do agentic stuff" innards of Octo into a reusable libocto, this does a reasonable-sized refactor of Octo's guts, primarily with the goal of moving everything other than trajectories into a generic libocto that doesn't depend on Octo-specific stuff. We already had moved compilers and tool factories to libocto; this moves autocompaction (and related compiler code)! Overall the changes are:

1. Move quota parsing outside of libocto, since it's very Synthetic-specific. Instead, libocto compilers return headers from all responses, and callers can parse the headers if they choose to (since quotas for syn are headers). We then of course in Octo parse the headers into quota data.
2. Make JSON autofixing optional in the compilers, since not all callers will necessarily want to use a JSON autofix model
3. Make rate limits and billing errors in the compilers be returned as Results instead of using try/catch with special errors. I believe this is the final place we were using `throw` for flow control inside libocto, and we're now fully migrated to Results for flow control. While in here, I also added support for OpenAI's billing/rate limit errors in the Responses compiler.
4. Remove some redundant ancient stuff from the standard compiler
5. Move compaction inside `libocto`. I also did a somewhat sizeable refactor to make compaction *no longer claim to generate tool tokens*: it already in practice actually couldn't generate tool tokens, but the TypeScript type system didn't know about that, and so callers always had to handle compaction as if it could possible be calling tools. This made no sense, and so compilers are now aware of whether or not tools are passed in and you no longer have to handle `type === "tool"` in your `onTokens` callbacks for autocompaction (or any other use case where Octo can't call tools).

This PR also fixes a longstanding bug in how tokens were counted in the compilers and autocompact, since I was in the autocompaction/compiler code anyway. Previously, we attempted to calculate the delta tokens in the compilers by examining the totals returned from the underlying SDKs, summing all the previous calculated deltas, and subtracting the sum(previous) from the total. However: *this is subtly broken*. Inference engines are *not* guaranteed to retokenize same-prefix message lists identically, and indeed in many cases do not: for example, DeepSeek and many other models will drop certain historical reasoning traces once they've passed certain windows, and thus the current total might be smaller than the previous total and it would appear like you had negative token messages. To fix this we simply stop attempting to calculate deltas, and always just store the latest total in the LLM IR. FYI @sdjidjev this is a breaking change to LlmIR, so it might be worth clearing out your old sessions if/when you rebase/merge your session code (since we haven't released session storage IMO no need to actually bump the version string in the stored session IRs).

This is the final boss before moving trajectory arcs inside libocto, which is itself a boss, so let's uh call this the final miniboss before migrating trajectory arcs. Once trajectory arcs are in, we're unblocked from working on Pi-like full trajectories that have callbacks for running tools + deciding permission questions + serializing state (the real final boss), and once that's done we could simply add this to Synthetic's web UI and have agentic, multi-turn, multi-tool behavior with full autocompaction, etc, using Octo's battle-tested guts.


source/agent/trajectory-arc.ts
```diff
@@ -232,14 +247,13 @@ export async function trajectoryArc({
...
-      reason: {
-        type: "request-error",
-        requestError: result.error.requestError,
-        curl: result.error.curl,
-      },
+      reason: compilerErrorToFinishReason(result.error),
     };
   }
 
+  const quota = parseQuotaFromHeaders(result.data.headers);
```
sdjidjev: this might need to be moved up before the early returns

sdjidjev (approved):
Nice refactor! Just one small comment.

source/agent/trajectory-arc.ts
reissbaker: fixed, good catch!

PR #214, opened 2026-06-16
Instruct Octo to avoid chunking

reissbaker:
Minor annoyance with the new offset and limit options for file reads: LLMs *love* to use these, way too much, and miss out on important context because they're trying to shrink the amount they're reading.

This tells them not to do that.


sdjidjev (approved):
omg yes, i've noticed this before.

PR #215, opened 2026-06-17
Allow outdated file writes

reissbaker:
Minor annoyance we talked about previously: Octo is very picky about forcing up-to-date full-file reads before allowing writes. Other coding agent harnesses don't seem to mind this as much, and e.g. when working with Codex, it's fine if I'm editing a file at the same time as Codex is. This has gotten especially annoying with partial file reads, since partial file reads aren't considered to un-dirty the file (since the change might not make its way into Octo if only part of the file was read).

This changes the file edit tool to allow writes to files that have been modified, and only returns the stale-file error message if the file is outdated *and* the edit can't be applied (e.g. because of a search string mismatch caused by an unseen-to-Octo file change).

Since appends and prepends can't have failed edits due to content changes (they only can fail if the entire file is deleted, which we catch separately), this means appends and prepends no longer check file staleness at all.

### DIFF
--- a/source/libocto/result.test.ts
+++ b/source/libocto/result.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from "vitest";
-import { err, flatten, ok } from "./result.ts";
+import { err, flatten, ok, unwrap } from "./result.ts";
 
 describe("flatten", () => {
   it("leaves a flat ok alone", () => {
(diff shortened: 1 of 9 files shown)


sdjidjev (approved):
LGTM! Down to remove append/prepend in this/another PR as we discussed in-person to reduce # of tools!

PR #217, opened 2026-06-17
Codex OAuth

reissbaker:
Mostly written by clanker, but I've reviewed it and cleaned up most of the LLM smells and I think this is reviewable.

The way Codex OAuth works is — crazily — spoofing the Codex app's client ID. Everyone does this and it seems generally allowed, e.g. OpenCode does this and apparently OpenAI is fine with it. Other than that most of this PR is:

1. Generalizing our auth code + auth UI flows to handle OAuth, in addition to standard `[API key, base URL]` tuple. This is like 99% of the PR. As part of this we add a new `~/.config/octofriend/oauth.json5` file, similar to the `keys.json5` file, to keep auth tokens out of configs (which might be synced via dotfiles).
2. A very very tiny update to make the subscription work. It uses a special `chatgpt.com` base URL and some special headers, and otherwise is just a normal OpenAI Responses client.

Auth flow:
image

image

It works:
image

I tested this and it seems to work fine. It even works with model switching back and forth, e.g. having Kimi generate some code and then swapping to 5.5 afterwards and having it work on the rest.

I also tested deleting the `oauth.json5` file, and it correctly handles that case: it auto-opens the Codex OAuth screen if you add gpt-5.5; this means that we don't throw crazy errors if you sync your `octofriend.json5` via dotfiles but don't sync your keys (which you shouldn't sync).


source/components/add-model-flow.tsx
```diff
@@ -426,15 +561,26 @@ function Context(props: FullFlowRouteData["context"] & Pick<Transitions<number>,
...
+
           done({
             baseUrl,
             model,
             nickname,
             context,
-            auth,
-          })
-        }
+            ...(auth ? { auth } : {}),
```
sdjidjev: is this doing something special? i feel like you could just leave it as `auth` even it's undefined, right?

source/preflight-auth.tsx
```diff
@@ -8,19 +8,32 @@ import {
...
 
 function resolveModelFromConfig(
   config: Config,
   model: Config["models"][number],
 ): Config["models"][number] {
-  const exact = config.models.find(
-    candidate => candidate.nickname === model.nickname && candidate.baseUrl === model.baseUrl,
-  );
+  const exact = config.models.find(candidate => {
+    if (model.type === "codex") {
```
sdjidjev: there are a few places doing this same matching logic, could be extracted to a helper!

sdjidjev (approved):
just tested out and it worked smoothly, this is awesome!!
pr lgtm with a few style comments but also have a question about a possible restructure to octofriend.json5. since codex is technicallyyy not a model it's a little confusing to have it referenced in the code as like `model.type = codex` and also in `octofriend.json5` i see it adds:
```
      type: 'codex',
      auth: {
        type: 'codex',
      },
```
and the duplication got me wondering if authentication deets should be stored under `models` or not 🤔 
like if OpenAI ever allowed API key for subscription users how would that look if you add that with the Codex OAuth

or if we ever had a flow where we allowed signing in with device code, how would we store that info

reissbaker:
Good notes on style!

Re: `type: "codex"`, the `type` field represents *what type of API the model uses*. Since Codex uses its own API type (a modified Responses API with a specific backend + different auth header scheme + specific copied client ID), that's why it's listed as its own type. For example, if `gpt-5.6` comes out, we could add it as a `codex` type model and use it via Codex subscriptions.

Alternatively, OpenAI also supports the "standard" Responses API, where it uses API keys + ordinary Bearer auth headers using `type: "responses"`. If OpenAI ever allows ordinary API keys to be used with subscriptions, that's how it'd work: you'd set the model to `type: "responses"` and it would require you to supply an API key, and it'd Just Work.

In terms of Synthetic allowing signing in via device code, there are a few ways we could do it:

1. Set up a true OAuth flow. We'd probably add a `type: "oauth"`, since we'd presumably use an Octo client ID as per ordinary OAuth rather than cloning Codex's.
2. Or, just do a simple `wandb` or `huggingface login` style flow, where the server generates a web page for you to log in with and then, once login succeeds, sends down the API key over a websocket or HTTP Streaming response. Then we'd keep things as `type: "standard"` and would just have a custom login flow for Synthetic in order to acquire the API key (if one doesn't exist yet).

source/preflight-auth.tsx
reissbaker: done!

source/components/add-model-flow.tsx
reissbaker: cleaned up!

PR #218, opened 2026-06-22
Add Session Resume DB schema

sdjidjev:
The octo Session Resume PR was getting pretty big so until we have stacked commits I wanted to at least separate a little and ask for feedback on just the DB side of things first.

This Schema supports a tree structured Session, where branches happen if there's two concurrent tabs open on the same session, or if we add a forking feature later.

Usually though, the "tree" will be more of a linked list.


source/session-history/schema/session-history-schema.ts
```diff
@@ -0,0 +1,155 @@
...
+    treeId: integer()
+      .notNull()
+      .references(() => trees.id, { onDelete: "restrict" }),
+    parentId: integer(),
+    launchId: integer()
+      .notNull()
+      .references(() => launches.id, { onDelete: "restrict" }),
+  },
+  table => [
+    uniqueIndex("tree_nodes_history_item_id_unique").on(table.historyItemId),
```
reissbaker: Technically in Drizzle I think you can remove these index names and Drizzle will autogenerate them, e.g.

`uniqueIndex().on(table.historyItemId)`

I think that's probably more readable

source/session-history/schema/session-history-schema.ts
```diff
@@ -0,0 +1,155 @@
...
+export const llmIrs = sqliteTable("llm_irs", {
+  id: integer().primaryKey({ autoIncrement: true }),
+  json: text().notNull(),
+});
+
+export const historyItems = sqliteTable(
+  "history_items",
+  {
+    id: integer().primaryKey({ autoIncrement: true }),
+    requestFailedId: integer().references(() => requestFailedItems.id, { onDelete: "restrict" }),
```
reissbaker, line 83: I think it would actually be nicer to do the opposite: `onDelete: "cascade"`, so that if you delete this row the corresponding data is also deleted. Ditto for all of the other `restrict` cases. Ultimately it would be nice to be able to just `DELETE <tree>` and the data in all corresponding tables cascades deletes, so you can easily just delete a conversation and have everything auto-cleaned-up.

source/session-history/schema/session-history-schema.ts
```diff
@@ -0,0 +1,155 @@
...
+  {
+    id: integer().primaryKey({ autoIncrement: true }),
+    kind: text({ enum: ["connect", "run"] }).notNull(),
+    containerTarget: text(),
+    dockerRunArgsJson: text(),
+    config: text(),
+    unchained: integer({ mode: "boolean" }).notNull(),
+  },
+  table => [
+    check("docker_launches_unchained_check", sql`${table.unchained} IN (0, 1)`),
```
reissbaker: Since Drizzle already has `mode: "boolean"` we don't need this check: the TypeScript compiler will handle this.

source/session-history/schema/session-history-schema.ts
```diff
@@ -0,0 +1,155 @@
...
+);
+
+export const localLaunches = sqliteTable(
+  "local_launches",
+  {
+    id: integer().primaryKey({ autoIncrement: true }),
+    config: text(),
+    unchained: integer({ mode: "boolean" }).notNull(),
+  },
+  table => [check("local_launches_unchained_check", sql`${table.unchained} IN (0, 1)`)],
```
reissbaker: Ditto on this boolean check

source/session-history/schema/session-history-schema.ts
```diff
@@ -0,0 +1,155 @@
...
+      .notNull()
+      .references(() => trees.id, { onDelete: "restrict" }),
+    parentId: integer(),
+    launchId: integer()
+      .notNull()
+      .references(() => launches.id, { onDelete: "restrict" }),
+  },
+  table => [
+    uniqueIndex("tree_nodes_history_item_id_unique").on(table.historyItemId),
+    uniqueIndex("tree_nodes_id_tree_id_unique").on(table.id, table.treeId),
```
reissbaker: I'm not sure that this unique index is enforcing anything? Since `id` is autoincrementing, every pair will be guaranteed to be unique just by virtue of the table ID being unique.

source/session-history/schema/session-history-schema.ts
```diff
@@ -0,0 +1,155 @@
...
+    uniqueIndex("tree_nodes_one_root_unique")
+      .on(table.treeId)
+      .where(sql`${table.parentId} IS NULL`),
+    index("tree_nodes_tree_id_idx").on(table.treeId, table.id),
+    index("tree_nodes_parent_tree_idx").on(table.parentId, table.treeId),
+    foreignKey({
+      name: "tree_nodes_parent_tree_fk",
+      columns: [table.parentId, table.treeId],
+      foreignColumns: [table.id, table.treeId],
+    }).onDelete("no action"),
```
reissbaker: Huh — what's the purpose of adding the foreign key here if we're taking no action on it?

source/session-history/schema/session-history-schema.ts
```diff
@@ -0,0 +1,155 @@
...
+      .notNull()
+      .references(() => launches.id, { onDelete: "restrict" }),
+  },
+  table => [
+    uniqueIndex("tree_nodes_history_item_id_unique").on(table.historyItemId),
+    uniqueIndex("tree_nodes_id_tree_id_unique").on(table.id, table.treeId),
+    uniqueIndex("tree_nodes_one_root_unique")
+      .on(table.treeId)
+      .where(sql`${table.parentId} IS NULL`),
+    index("tree_nodes_tree_id_idx").on(table.treeId, table.id),
```
reissbaker: What's the purpose of the compound index here? If we already have the `id`, it should be a simple index lookup to find the row.

source/session-history/schema/session-history-schema.ts
```diff
@@ -0,0 +1,155 @@
...
+    index("tree_nodes_parent_tree_idx").on(table.parentId, table.treeId),
```
reissbaker: I might leave out these plain indexes for now — without the accompanying code, it's hard to see what query patterns these reflect. We can always add the indexes alongside the code later if they're needed. My sense is we probably aren't querying on `parentID` very often, and are more likely just loading the full tree and doing the rest of the tree-walk in memory, so all we'd need here is `index().on(table.treeId)`

reissbaker (approved):
Sensible table structure! Left some nits on code style (in Drizzle I believe you can leave out the index name strings, and I think that's more readable), removing some indexes + checks, and swapping to `onDelete: "cascade"` which I think will be more useful for us longer term.

drizzle/0002_silver_circuit.sql
```diff
@@ -94,6 +96,33 @@ CREATE TABLE `trees` (
...
 CREATE INDEX `trees_updated_at_idx` ON `trees` (`updated_at`);--> statement-breakpoint
+CREATE TRIGGER `tree_nodes_require_new_leaf`
+BEFORE INSERT ON `tree_nodes`
+WHEN NEW.`is_leaf` <> 1
+BEGIN
+	SELECT RAISE(ABORT, 'new tree nodes must be leaves');
+END;
+--> statement-breakpoint
+CREATE TRIGGER `tree_nodes_mark_parent_non_leaf`
+AFTER INSERT ON `tree_nodes`
```
reissbaker: Might as well add an `AFTER UPDATE` version too, although it'll probably not be used much.

reissbaker (commented):
Oh, one more comment on code style: I would separate your handwritten SQL (the triggers) from the Drizzle autogenerated ones, so it's clear which migrations are autogenerated and don't need to be read to understand the code, vs which ones are custom ones. You can add custom migrations in Drizzle Kit with `generate --custom`: https://orm.drizzle.team/docs/kit-custom-migrations

source/session-history/schema/session-history-schema.ts
sdjidjev: oh the idea was to prevent nodes that have children from being deleted. (as a safeguard)
maybe changing it to say `restrict` instead of `no action` is more explicit though!

the foreign key also makes sure that a node's parentId is a node in the same tree, so figure it's worth keeping! i'll also add a comment and rename it since it is a little hard to parse what this is doing at a glance 😅

sdjidjev: oh wait i just read your cascade comment! lemme see if/how i should keep the fk here based on the cascade idea 🤔

source/session-history/schema/session-history-schema.ts
sdjidjev: oh yeah good call, way cleaner!
i left `tree_nodes_one_root_unique` as a `uniqueIndex`, since `unique` doesn't have a `.where()` method (if i end up using it...still doing a pass to make sure i don't over-unique things)

i wish `index(` had a similar feature but i think i have to still manually name all indexes.

source/session-history/schema/session-history-schema.ts
sdjidjev: ah ok, makes sense: removed!

source/session-history/schema/session-history-schema.ts
sdjidjev: Sounds good! I'll add them in as needed in future PRs.

source/session-history/schema/session-history-schema.ts
sdjidjev: Definitely agree to `onDelete: "cascade"` for things like a tree id and a node id, but thinking it's best to leave it as `restrict` for payload type things like a `history_item.` Since we should never independently delete a payload (should be deleting at the node level only), seems fair to restrict that behavior.
Lmk if that makes sense!

reissbaker: Makes sense, good point!

source/session-history/schema/session-history-schema.ts
sdjidjev: For this index specifically, the foreign key defined here won't work. 
```
    foreignKey({
      name: "tree_nodes_parent_same_tree_fk",
      columns: [table.parentId, table.treeId],
      foreignColumns: [table.id, table.treeId],
    }).onDelete("cascade"),
```
if i try deleting `unique().on(table.id, table.treeId)` SQL says `"foreign key mismatch - "tree_nodes" referencing "tree_nodes"` (which is an unhelpful message lol).....but it's cuz SQL is silly and doesn't realize that since `id` is autoincrementing, the combo of `id`+`tree_id` is guaranteed unique. You have to explicitly write that it will be unique for SQL to allow the foreign key to exist.

sdjidjev (commented):
@reissbaker thanks for your review!
Updated the PR with the feedback, PTAL to make sure the changes make sense!
Also, some of your comments on the updated files disappeared on the `Files changed` view but I replied to some of them with some of my reasoning as well!

reissbaker (approved):
LGTM! :ship:

PR #219, opened 2026-06-23
Remove append, prepend tools

reissbaker:
As previously discussed — these were tools added back in the dark ages when LLMs often severely struggled to edit files correctly, and they could at least fall back to appending/prepending to files without needing to match search strings. They're basically just tool bloat now.


sdjidjev (approved):
👏

PR #220, opened 2026-06-29
Add Session Resume to Octo

sdjidjev:
Sessions can be resumed now with command `octo --resume <session_id>`. Works with either local or docker launches.

The session preserves all flags & options from the time you last started (or resumed) that session.

We store the history as a tree:
```
trees (session id is tree.name)
│
└── tree_nodes (the conversation)
    │
    id=1  "hello octo"  ← root (first message)
    │
    └── id=2  "🐙 hi there, steph, i am octo"    ← assistant reply
        │
        ├── id=5  "octo, i love that you can resume"  
        │   │
        │   └── id=6  "🐙 yea, it's pretty great"      ← highest id leaf 
        |                                               (the session would be resumed here)
        │
        └── id=3  "yo octo, i just resumed this session in a different terminal tab"   ← sibling branch 
            │
            └──  id=4  "🐙 sg, if the other tab is updated more recently,
                          this sibling branch won't show up when resumed."  
```
Each node points to one history_item, which points to one payload (llm-ir, notification, etc.). New inserts are always leaves; inserting under a node un-leaves the parent. Resume will walk from the highest-id leaf back to the root. (node id's autoincrement).

Screenshot 2026-07-10 at 6 16 34 PM


source/cli/cli.tsx
```diff
@@ -8,87 +8,122 @@ import os from "os";
...
-        config: opts.config,
-        unchained: opts.unchained,
         transport,
+        sessionRecord: null,
+        parsedCliArgs: {
+          kind: "docker-connect",
+          target,
+          config: opts.config,
+          unchained: opts.unchained,
+        } as DockerConnectCommand,
```
reissbaker: Why the cast here?

source/cli/cli.tsx
```diff
@@ -98,18 +133,108 @@ docker
...
+): Promise<{
+  sessionRecord: SessionRecord;
+  transport: Transport;
+  parsedCliArgs: ParsedCliArgs;
+} | null> {
+  const sessionRecord = await loadSessionRecord(resumeSessionId);
+  if (sessionRecord == null) {
+    console.error(`No session found with ID ${resumeSessionId}.`);
+    process.exitCode = 1;
+    return null;
```
reissbaker: Rather than modifying the global state and then needing the caller to understand they need to exit, I think `process.exit(1)` is probably the right call here

source/session-history/index.ts
```diff
@@ -0,0 +1,504 @@
...
+  cliArgs: ParsedCliArgs,
+): SessionRecord {
+  return {
+    metadata: { sessionId, cwd, transportKind, cliArgs },
+    nodePath: [],
+  };
+}
+
+export async function loadSessionRecord(sessionId: string): Promise<SessionRecord | null> {
+  return runWriteTransaction(tx => {
```
reissbaker: This probably doesn't need to be in a transaction, and I think it doesn't need to hit the DB as frequently as it's doing. The simplest approach would be:

1. Load the entire tree in-memory (as you're doing below), but also join it against the various subtables using Drizzle query's `with` field: https://orm.drizzle.team/docs/rqb#include-relations
2. In-memory, walk the tree to find the most recent leaf (rather than hitting the DB again)
3. In-memory, load the node path to the leaf (rather than doing N queries for each node in the path)

source/session-history/index.ts
```diff
@@ -0,0 +1,504 @@
...
+
+export class SessionHistory {
+  private activeSession: SessionRecord;
+  private launchId: number | null = null;
+  private pendingSave: Promise<boolean> = Promise.resolve(false);
+  constructor(sessionRecord: SessionRecord) {
+    this.activeSession = sessionRecord;
+  }
+
+  private enqueue(operation: () => Promise<boolean>): Promise<boolean> {
```
reissbaker: Is this necessary? Doesn't better-sqlite3/Drizzle already handle write concurrency?

source/state.ts
```diff
@@ -689,3 +693,11 @@ export function useModel() {
 
   return getModelFromConfig(config, modelOverride);
 }
+
+export function subscribeSessionHistory() {
+  return useAppStore.subscribe((state, prevState) => {
+    if (state.history !== prevState.history && state.sessionHistory != null) {
+      state.sessionHistory.replace(state.history);
```
reissbaker: I think a lot of the complexity of this PR is coming from this. It's treating sessions as a post-hoc syncing problem rather than first-class living in the SQLite database.

I think this should probably be restructured. We shouldn't create history items in memory and then try to sync them afterwards to the DB; instead, we create history items directly in the DB immediately, and no syncing is ever required. That removes the necessity of walking the history and finding prefixes, and is also concurrency-safe since you're not rewriting large swaths of history at once. It removes the need for transactions, since you're just writing nodes with an `insert(table).values({ ... })`. There shouldn't be a `SessionHistory` sync manager; instead, `state.ts` just writes the nodes right away.

The only complex history code is in the read path, not the write path: we need to read the tree, and then walk the path from the latest leaf in-memory so that we have a linear history to present to the initial `state.ts` init code. But I think the write path can be streamlined by replacing all the code that creates history items in memory with insert calls that create the history items in sqlite instead, and use [`.returning()`](https://orm.drizzle.team/docs/sqlite/insert#returning) to get the object in memory.

reissbaker (changes requested):
This is a good start but I think there's a lot of complexity stemming from one decision: to largely avoid changing `state.ts`, and to instead try to subscribe to events from the Zustand state and sync them to the DB.

I think this code would be simpler and smaller if we instead *performed the writes directly in Zustand* and didn't try to create in-memory HistoryItems ever: instead, history items are only ever created by inserts with Drizzle's [`.returning()`](https://orm.drizzle.team/docs/sqlite/insert#returning) syntax. This would eliminate almost all of the code in the write path, I think, and leave only the read path as requiring some ceremony.

The read code could use some optimization/simplification as well; I think the primary change there is to just read the entire tree into memory once and operate on it, rather than making lots of sub-queries. You already have the tree loaded into memory; you might as well use it rather than re-reading data from the table!

Re: deleting and listing sessions — I think that's probably worth doing in a followup PR, rather than making this one larger.

source/state.ts
```diff
@@ -152,6 +172,12 @@ export const useAppStore = create<UiState>((set, get) => ({
   query: "",
   clearNonce: 0,
   lastUserPromptIndex: null,
+  _session: null,
+  getSession: () => {
+    const session = get()._session;
+    if (session == null) throw new Error("Session is not initialized.");
```
sdjidjev: @reissbaker not sure if this is the best way to set this up, but i found it annoying to have a nullable `session` property (`session`'s id can be nullable but the `metadata` field is required)
is there a better/preferred way to initialize `session` in `cli.tsx` than what i had with calling `startNewSession` and  `hydrateSession`?

source/session-history/index.ts
```diff
@@ -0,0 +1,316 @@
...
+export function loadSession(sessionId: string): LoadedSession | null {
+  const tree = loadSessionTree(sessionId);
+  if (tree == null) return null;
+
+  let mostRecentLeaf: SessionTreeNode | undefined;
+  const nodesById = new Map<number, SessionTreeNode>();
+  for (const node of tree.nodes) {
+    nodesById.set(node.id, node);
+    if (node.isLeaf && (mostRecentLeaf == null || node.id > mostRecentLeaf.id)) {
+      // todo should we add a created at field to the tree nodes to find most recent leaf instead of relying on the autoincrementing id?
```
sdjidjev: i feel like for now keeping autoincrementing id check is fine for seeing what was most recently created, but maybe when forking sessions is a thing we could add a created_at field?

source/state.ts
```diff
@@ -103,9 +110,11 @@ export type UiState = {
   quotaData: QuotaData | null;
   byteCount: number;
   query: string;
-  history: Array<HistoryItem>;
+  readonly history: readonly HistoryNode[];
```
sdjidjev, line 114: not sure of the best way to enforce that the state's `history` should not be modified without going through the db.

source/state.ts
```diff
@@ -131,11 +140,22 @@ export type UiState = {
...
   _maybeHandleAbort: (signal: AbortSignal) => boolean;
   runAgent: (args: RunArgs) => Promise<void>;
 };
 
+function appendAndPersistHistory(
+  session: Session,
+  prevHistory: readonly HistoryNode[],
+  itemsToInsert: HistoryItem[],
+): HistoryNode[] {
+  if (useAppStore.getState().getSession() !== session) return [...prevHistory];
```
reissbaker: This fallback behavior seems like a silent bug: it's dropping history items on the floor if the state has somehow become corrupt, which further corrupts Octo's internal state rather than fixing it. As far as I can tell, this codepath should never happen, and if it does happen it's unrecoverable; for those cases we should probably just hard error like: `throw new Error("Session desynchronized; this is an error in Octo")`

source/state.ts
reissbaker: Ah, I think I see the issue.

You're making `_session` nullable, and expect that the App will either hydrate it at runtime or create a new one at runtime. The reason it's nullable is because when you create the store, there's no way to know a) whether to read or create a session, and b) which session to read (if you need to read a session).

I'd probably fix this by just taking `_session` out of Zustand's state, and requiring it to be passed in. For example, this is how we treat `transport`: rather than putting it in the Zustand state (which needs to be initialized, and we don't have a valid thing to initialize it with until later), we just pass it around as an argument.

For `createNewSession`, it can still do all of the ordinary Zustand-state-updating; it just also returns a new Session object that callers are expected to use, rather than updating its internal `_session`.

source/session-history/index.ts
reissbaker: For forking sessions, auto-incremented IDs should still work, since the later it was created the higher the ID will be. There's no ordering difference between using auto-incrementing IDs and created-at fields, since the earlier created-at row will also have a lower ID.

That being said — it might be nice to add a `createdAt` field now! It's impossible to backfill, since we just simply don't know when an item was created if we don't store it somewhere. And I could imagine it could be useful to show timestamps in the UI someday; if we don't store it now, we'll never be able to do that.

source/session-history/index.ts
```diff
@@ -0,0 +1,316 @@
...
+          parentId: currParentId,
+          isLeaf: true,
+          launchId,
+        })
+        .returning({ id: treeNodes.id })
+        .get();
+      insertedNodes.push({ ...historyItem, nodeId: insertedTreeNode.id });
+      currParentId = insertedTreeNode.id;
+    }
+    tx.update(trees).set({ updatedAt: Date.now() }).where(eq(trees.id, treeId)).run();
```
reissbaker, line 209: I believe you can use a built-in Drizzle helper for this! 
https://orm.drizzle.team/docs/latest-releases/drizzle-orm-v0305

reissbaker (approved):
Left a few nits but generally LGTM!

I think we're starting to outgrow what feels natural to put in Zustand, especially with e.g. history living in Zustand and therefore awkward session stuff (since either sessions have to be nullable, or they have to live outside of Zustand).

That being said, I think some libocto refactors I have planned will mostly resolve this (by moving a lot of state outside of Zustand and into libocto). So, we can keep this mostly as-is for now and we'll resolve it later.

source/session-history/index.ts
sdjidjev: yeah sg!! you're right that if we add it later we could miss out on past sessions having that info

source/state.ts
sdjidjev: ohhh yes smart! sounds good.

source/state.ts
sdjidjev: ooh ok! i was nervous that throwing would be disruptive but hopefully it would basically never happen and thus a non-issue 🙏

reissbaker: Agreed! Also, with session save/resume, you don't lose toooo much if this ever happens since you can always resume the session :)

sdjidjev (commented):
@reissbaker thanks for your comments!
Addressed your comments; the biggest change was moving `session` to be passed in to `state` rather than making zustand keep track of it. 

Also now starting a new session through the Octo menu will clear the old history and print out a resume hint saying how to resume the previous session! 

Image

I believe I addressed all your comments, although now I'm experiencing the bug where past comments/replies are disappearing on me, lmk if I missed something!

source/cli/cli.tsx
```diff
@@ -92,18 +118,95 @@ docker
...
+          unchained: opts.unchained,
+        },
       });
     } finally {
       await transport.close();
     }
   });
 
-async function runMain(opts: { config?: string; unchained?: boolean; transport: Transport }) {
+async function buildSessionContext(
```
fireph: Feel like there is a better name for this function since it is doing more than building session context.

source/ir/convert-history-ir.ts
```diff
@@ -1,6 +1,6 @@
-import type { HistoryItem } from "../history.ts";
 import type { OctoIR } from "./octo-ir.ts";
 import type { TrajectoryOutputIR } from "../agent/trajectory-arc.ts";
+import { HistoryItem } from "../session-history/index.ts";
```
fireph: `import type`

reissbaker:
LGTM! I think there's one tiny missing nit around using Drizzle's automatic `updatedAt` helper (now missing from the main Github review page, :scream_cat: ) but I'll merge this and add it in main to unblock :)

PR #221, opened 2026-07-01
Add a 'write curl file' alongside the 'copy curl command' option

billycao:
I can't use terminal clipboards since I ssh into a devserver. (I actually use a workaround so I *can* clipboard from my devserver to my macbook, but it doesn't use what octo uses (`xsel`), which requires X server forwarding to be set up :P)

This PR is a small proposal to add a "write curl file" option on errors.

Tested locally:
CleanShot 2026-06-30 at 21 59 53@2x
CleanShot 2026-06-30 at 22 03 36@2x


reissbaker (approved):
LGTM!

PR #224, opened 2026-07-14
Paintcannon port

reissbaker:
This is a mostly-mechanical port of Ink's React to Paintcannon's React, done by gpt-5.6-sol and reviewed by me (sometimes with feedback and changes). It's a lot of changes, but they should almost entirely be "turn this Box/Text styled via props into a Div/Span styled with CSS," or equivalent e.g. swapping to Paintcannon's paste events API, using its DOM-like keyboard event bubbling instead of Ink's `useInput` hook, etc.

I also added a couple quality-of-life niceties made possible via Paintcannon:
- Octo now auto-dims when it doesn't have focus
- The Octo Friend logo text is centered on boot
- Octo has a subtle dark blue background color now, so we're less at the mercy of random people's terminal settings in terms of readability

image

Some of this port was done by having gpt-5.6-sol write an AST parser and rewriter, and auto-rewriting `tsx` files to change `Box` and `Text` components into `Div` and `Span` components. One downside of this is that there were some whitespace changes that came along with the AST rewriter. However, generally speaking this port is fairly straightforward mechanically.

And, it works really really really well.

I wasn't planning on doing this until a few other features shipped. But: the Session Resume PR somehow made Octo+Ink's existing flicker problem inside tmux much worse, to the point of being more or less unusable. I suspect this is because better-sqlite3 is synchronous, which they claim results in improved throughput. That throughput claim may be true, but I suspect it *also* results in tons of flicker inside Ink if you're using it inside tmux or other programs that don't know about the (relatively recent) terminal sync APIs that are meant to combat this; without those vsync-equivalents, even milliseconds of blocking the event loop will cause partial, torn paints when the entire paint system is written in JS and it's getting its event loop blocked by SQLite. This happened pretty painfully to both me and @billycao , implying the problem is indeed something about torn paints inside tmux, since we both are tmux users but on vastly different setups otherwise.

Switching to Paintcannon fixes this entirely, because:

1. JS uses NAPI-RS to send messages to Paintcannon's Rust code
2. Paintcannon sends those messages over a channel to a separate Rust rendering thread
3. The separate thread owns the writes to stdout, so the JS event loop blocking is meaningless: Rust will never block due to JS, since it's running on a different thread, and so paints won't get torn as a result of the event loop getting blocked

Also, Paintcannon exposes a transaction API that allows clients to tell Paintcannon they're in the middle of making rendering changes, and haven't finished yet, so Paintcannon should just buffer and wait before painting, which especially helps reduce screen tearing. And `paintcannon-react` automatically makes use of that API, so there's pretty much zero flicker ever, even inside tmux.

(Paintcannon also uses the terminal sync protocol that many modern terminals use, and Ink uses, to avoid flicker, but tmux doesn't understand it and so it doesn't save you in tmux.)

Typing also feels way better in this port, since it's done mostly-natively. And of course:

* Scrollbars Just Work (including nested scroll, so it's for the first time ever possible to scroll the assistant thinking data)
* Paste events, including image drag and drop, Just Works using browser APIs. Paintcannon actually copied Octo's code for this, but it means we can move the code out of Octo at least.
* Text selection works in alt screen mode.
* Vim multiline cursor movement works significantly better, since Paintcannon exposes cursor movement APIs that Ink doesn't.
* Octo probably works better over SSH
* Borders don't get randomly corrupted if you change terminal window sizes
* Selecting text inside bordered content doesn't select the borders in alt-screen mode with mouse capture, so copy/paste works better
* Word wrapping works reliably
* No more need for "static" content
* Historical entries resize correctly on terminal window size changes, since they're not static
* etc etc

I ended up needing to add some features and fix a few bugs in Paintcannon as a result of doing this port, since Octo is a pretty complex beast at this point. But: it works now! And there's zero flicker!


source/components/file-suggestions/suggestion-list.tsx
```diff
@@ -1,33 +1,47 @@
...
+            style={{
+              display: "flex",
+              whiteSpace: "pre-wrap",
+            }}
+          >
             {isSelected ? (
-              <Text inverse>
+              <Span
+                style={{
+                  color: "#111827",
```
fireph: might be nice to avoid arbitrary colors and use theme.ts

source/components/selection/select-input.tsx
```diff
@@ -89,98 +86,97 @@ function SelectInput<V>({
...
-
         return (
-          // @ts-expect-error - `key` can't be optional but `item.value` is generic T
-          <Box key={item.key ?? item.value}>
-            {React.createElement(indicatorComponent, { isSelected })}
-            {React.createElement(itemComponent, { ...item, isSelected })}
-          </Box>
+          <Div
+            key={item.key ?? String(item.value)}
+            style={{
```
fireph: could most of the <Div be <FlexDiv or something and have
```
display: "flex",
whiteSpace: "pre-wrap",
```
be the default on it?

source/components/add-model-flow.tsx
```diff
@@ -582,71 +779,102 @@ function Context(props: FullFlowRouteData["context"] & Pick<Transitions<number>,
...
-          <Text>
-            Format the number in k: for example, <Text color={color}>32k</Text> or,{" "}
-            <Text color={color}>64k</Text>.
-          </Text>
-        </Box>
+            </Span>
+          </Div>
+          <Span>
+            Format the number in k: for example,{" "}
+            <Span
```
fireph: nit: this might be a spot to manually edit since this multiline format is less readable than what was there before

source/components/app-focus-overlay.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+  return (
+    <Div
+      style={{
+        position: "absolute",
+        top: 0,
+        left: 0,
+        width: "100%",
+        height: "100%",
+        zIndex: 2147483647,
```
fireph: this might be good in a variable 😅

source/components/diff-renderer.tsx
```diff
@@ -204,7 +246,18 @@ function DiffSet({
...
       >
-        {oldRemoved ? <Text color="black"> - </Text> : <Text>{"  "}</Text>}
+        {oldRemoved ? (
+          <Span
+            style={{
+              color: "black",
+            }}
+          >
+            {" "}
+            -{" "}
```
fireph: nit

source/components/diff-renderer.tsx
```diff
@@ -215,12 +268,22 @@ function DiffSet({
...
       >
-        {newAdded ? <Text color="black"> + </Text> : <Text>{"  "}</Text>}
+        {newAdded ? (
+          <Span
+            style={{
+              color: "black",
+            }}
+          >
+            {" "}
+            +{" "}
```
fireph: nit

source/markdown/index.tsx
```diff
@@ -76,194 +78,362 @@ function TokenRenderer({ token }: { token: Token }): React.ReactElement {
...
-  return <Text>{token.text}</Text>;
+  return <Span>{token.text}</Span>;
 }
-
 function HeadingRenderer({ token }: { token: Tokens.Heading }) {
   const indent = Math.max(0, token.depth - 1) * 2; // Convert to padding units
 
-  const colors = ["magenta", "blue", "cyan", "green", "yellow", "red"] as const;
-
+  const colors = ["#d4d4d4", "#c8c8c8", "#bfc3c7", "#b9b8c0", "#b5b0bb", "#afa9b5"];
```
fireph: nit: maybe base colors in theme?

fireph (approved):
all looks reasonable to me, just some small nits 🥳

source/components/selection/select-input.tsx
reissbaker: Done! Swapped to `TerminalFlex` which means flex + pre-wrap

source/components/add-model-flow.tsx
reissbaker: Condensed to single-line spans :)

source/components/diff-renderer.tsx
reissbaker: Condensed to single-line for readability

source/components/diff-renderer.tsx
reissbaker: Condensed this too!

source/components/app-focus-overlay.tsx
reissbaker: Done!

source/components/file-suggestions/suggestion-list.tsx
reissbaker: Swapped these to import foreground color and bg color from theme.ts

source/markdown/index.tsx
reissbaker: Done!

PR #233, opened 2026-07-20
Fix aborted tool calls for Responses API

reissbaker:
Minor but annoying bug with session resume: if Octo crashes mid-tool-call, and you resume Octo later, it will perma-break OpenAI models using the Responses compiler. OpenAI expects *all* function calls to have function output, even if the output is just e.g. "Aborted." However, if Octo crashed and never generated a response, on resume we'll still have the same assistant requests stored in the DB, but we'll never insert function call outputs indicating that we didn't run the function.

Chat completions are generally a looser API: it's alright to not have a response to a function call. But apparently this isn't okay in the Responses API.

The ideal fix IMO is that we should resume any attempted function calls on session resume. But for now, this patches the perma-break for OpenAI models by auto-inserting "Aborted" function call outputs in the Responses compiler for any function calls that lack output.

Tested on a perma-broken session with GPT-5.6-Sol, and this unbroke the session.

### DIFF
--- a/source/libocto/compilers/responses.ts
+++ b/source/libocto/compilers/responses.ts
@@ -130,6 +130,13 @@ async function toResponseInput<A extends Agent<any, any, any>>(
     items.push(...responseInputFromIr(ir, modalities));
   }
 
+  /*
+   * The OpenAI Responses API mandates that all function calls *must* have function call outputs
+   * associated with them on requests, unlike other APIs which allow you to simply leave them out if
+   * you want to (e.g. if you didn't run them). To normalize this behavior, we insert "Aborted"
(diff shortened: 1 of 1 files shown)


fireph (approved):
lgtm! but since this is just a temporary fix until the session resume real fix is in, is it worth adding a todo here or a comment explaining why this code is needed?

reissbaker:
So, I think it's actually just useful to keep in place — even once the fix is in, since it's non-obvious from LLM IR that all functions must have outputs (esp since not all APIs have this requirement). But we should also do the fix where session resume also resumes tool calls IMO. I'll add a comment though explaining

PR #234, opened 2026-07-21
Improve grep tool

reissbaker:
I've noticed that GLM-5.2 often forgets to include a `pattern` for grep, and then subsequently gets annoyed by using the tool and swaps to shelling out to literal Unix `grep` via the `shell` tool. This is annoying, because the point of the `grep` tool is to avoid permissions prompts for safe grep usage in collaboration mode; using grep via `shell` always triggers permissions prompts in collaboration mode.

This makes the `pattern` field required for the grep tool. GLM-5.2 always uses it when it's required and seems to never forget. (And if it did forget, Structural would give a helpful error.)

This PR also updates Structural to the latest version, which includes much-improved error rendering. This should help LLMs recover from messed up tool calls more reliably (like `grep`), and also should help users who previously encountered inscrutable errors from accidentally-broken configs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -34,7 +34,7 @@
         "react": "^19.2.6",
         "shell-quote": "^1.8.3",
         "string-width": "^7.2.0",
-        "structural": "^0.0.29",
+        "structural": "^0.0.33",
         "to-rotated": "^1.0.0",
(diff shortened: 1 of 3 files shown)


sdjidjev (approved):
Sweet, glad it's fixed!

PR #237, opened 2026-07-22
Fix list spacing discrepancy

sdjidjev:
Before
Screenshot 2026-07-21 at 6 41 23 PM

After
Screenshot 2026-07-21 at 6 41 57 PM

Ink / Yoga's [default flexShrink is 0](https://github.com/react/yoga/blob/main/yoga/style/Style.h#L46) which was why this wasn't a problem before.
Was wondering whether Paintcannon should also have its default `flexShrink` set to `0`, but noticed that the default for `flexShrink` for **web** in Yoga is actually `1`. Since Paintcannon aims to be as close to regular CSS as possible, we should keep the default as `1`.


fireph (approved):
🥳

PR #243, opened 2026-07-24
Add octo session delete

sdjidjev:
With `octo session delete <session-id>` you now have the power to erase that session!
**After deletion**
Screenshot 2026-07-24 at 6 51 49 PM

**Behavior if you add a message to a session that had been deleted in another tab.**
Screenshot 2026-07-24 at 6 51 58 PM

### DIFF
--- a/source/app.tsx
+++ b/source/app.tsx
@@ -56,7 +56,8 @@ import { useShallow } from "zustand/react/shallow";
 import { KbShortcutPanel } from "./components/kb-select/kb-shortcut-panel.tsx";
 import { Item, ShortcutArray } from "./components/kb-select/kb-shortcut-select.tsx";
 import { useAppStore, RunArgs, useModel, InflightResponseType } from "./state.ts";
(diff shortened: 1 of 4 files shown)


source/app.tsx
```diff
@@ -630,6 +631,7 @@ function BottomBarContent({ inputHistory }: { inputHistory: InputHistory }) {
   const model = useModel();
   const transport = useContext(TransportContext);
   const session = useSession();
+  const showToast = useToast();
```
fireph, line 635: should this be `const toast = useToast();` for consistency with all the other custom hooks? I don't feel strongly about this though. `showToast` is kind of a better name for calling it so 🤷

sdjidjev: yeah i just named it that cuz it's technically a function as opposed to the other plain constants. 🤔 
i think `showToast` makes more sense when used lower down in the file.

fireph (approved):
Image

PR #246, opened 2026-07-24
Remove overlay in favor of background color + scrollbar dim

reissbaker:
Removes the new overlay-on-blur effect, since:

1. It made text hard to read when the app wasn't focused, and
2. It made scrolling impossible when the app wasn't focused (since the overlay was on top, it consumed the scroll events)

Instead it replaces it with just changing the background color and scrollbar color when the app is unfocused, which is still a nice touch for terminal pane abusers such as myself, but much less invasive.

Focused octo:
image

Unfocused octo:
image


fireph (approved):
looks so slick!

PR #250, opened 2026-07-28
Fix dangling tool calls

reissbaker:
Now that we have Kimi K3 on Synthetic, I pointed Kimi-in-Octo at Octo and told it to find bugs. Two of the more-critical bugs it found were:

1. When aborting a tool call, we left it dangling without a corresponding tool call output IR. The Responses compiler auto-fixes all of these, but the standard and Anthropic compilers don't. The standard chat completions API doesn't *break* if you leave dangling tool calls without outputs, but it is *weird* and probably out-of-domain for models. K3 claimed that Anthropic breaks, although I didn't check; regardless it's still a weird state for an LLM to end up in where we just ignore its tool calls.
2. Image-reading tool calls also left dangling tool calls. Rather than turning image reads into tool output IRs, we turned them into faked user message IRs.

This PR fixes both sources of dangling, unanswered tool calls.

### DIFF
--- a/source/app.tsx
+++ b/source/app.tsx
@@ -612,7 +612,7 @@ function BottomBarContent({ inputHistory }: { inputHistory: InputHistory }) {
         setVimMode("NORMAL");
         return;
       }
-      abortResponse();
+      abortResponse(session);
       if (modeData.mode === "menu") closeMenu();
     }
     if (event.ctrlKey && event.key === "p") {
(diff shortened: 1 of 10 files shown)


sdjidjev (approved):
This LGTM!
Either on this PR or another one, might be nice to also call `abortResponse(session)` in `ExitOnDoubleCtrlC` to cover that case as well (i shamelessly always use that to exit octo 🙈)

reissbaker:
Good point! Added that here too

PR #251, opened 2026-07-28
Preserve model used during session when resumed

sdjidjev:
With this fix, whichever model was used most recently when a session was exited, will be the same model used on resumption. We store the model config as a stringified versioned json in the DB. We first try an exact match and if that doesn't work, a fuzzy match with model name and the base url.

If the fuzzy match can't find a model in the `octofriend.json` config, it will default to `octo`'s default model.
Screenshot 2026-07-28 at 3 24 59 PM

### DIFF
--- a/drizzle/0006_history_item_model_json.sql
+++ b/drizzle/0006_history_item_model_json.sql
@@ -1,0 +1,1 @@
+ALTER TABLE `history_items` ADD `model_json` text;
--- a/drizzle/0006_history_item_model_json.sql
+++ b/drizzle/0006_history_item_model_json.sql
@@ -1,1 +1,7 @@
 ALTER TABLE `history_items` ADD `model_json` text;
+--> statement-breakpoint
+UPDATE `history_items`
+SET `model_json` = 'octo-no-model-recorded'
+WHERE `model_json` IS NULL;
+--> statement-breakpoint
+ALTER TABLE `history_items` ALTER `model_json` SET NOT NULL;
--- a/source/app.tsx
+++ b/source/app.tsx
@@ -18,7 +18,7 @@ import {
   ConfigContext,
   ConfigPathContext,
   SetConfigContext,
-  getModelFromConfig,
+  matchModelFromConfig,
   mergeEnvVar,
   readAuthForModel,
   useConfig,
@@ -58,6 +58,7 @@ import { Item, ShortcutArray } from "./components/kb-select/kb-shortcut-select.t
 import { useAppStore, RunArgs, useModel, InflightResponseType, nextToolAction } from "./state.ts";
 import { SessionNotFoundError } from "./session-history/index.ts";
 import type { HistoryNode, Session } from "./session-history/index.ts";
+import { tryDeserializeModelJson } from "./session-history/model-json.ts";
 import { Octo } from "./components/octo.tsx";
 import { Menu } from "./menu.tsx";
 import SelectInput from "./components/selection/select-input.tsx";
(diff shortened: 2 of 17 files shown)


source/session-history/schema/session-history-schema.ts
```diff
@@ -99,6 +99,7 @@ export const historyItems = sqliteTable(
   "history_items",
   {
     id: integer().primaryKey({ autoIncrement: true }),
+    modelNickname: text(),
```
sdjidjev: was thinking about whether to make this `notNull`....
thought i should leave it as nullable:
  - in case `historyItem` could somehow be model-less?
  - making up a default is kinda weird and would make the "no longer in config" check more messy

source/app.tsx
```diff
@@ -266,24 +266,44 @@ export default function App({
...
+  const showToast = useToast();
+  useEffect(() => {
+    if (modelOverride == null) return;
+    if (currConfig.models.some(model => model.nickname === modelOverride)) return;
+    showToast(
+      <Span style={{ color: "red" }}>
+        {`This session used the model "${modelOverride}", which is no longer in your config. Falling back to the default model.`}
+      </Span>,
+    );
+  }, [currConfig.models, modelOverride, sessionHydrationNonce, showToast]);
```
reissbaker: The `useEffect` here seems like it could trigger too many times. For example, if you add a model, this would trigger, since it would change `currConfig.models`.

Instead you could structure it like:

1. Load the current model outside of the `useEffect`.
2. Stick it in a ref
3. Have the useEffect depend on the ref, and on the hydration nonce
4. Access it from `ref.current` inside the `useEffect`

source/session-history/schema/session-history-schema.ts
reissbaker: Couple thoughts on this:

1. I think everywhere that inserts a history item currently has access to the current model. That implies to me this should be non-nullable, and the functions should stop accepting `string | null` and just accept `string`. Unless I'm missing a spot that doesn't have a model?
2. You can just call `.notNull()` without providing a default; that will just force Drizzle (and SQLite) to demand you pass in a value here rather than `null`
3. *But* I actually think `modelNickname` is not the right field to track. Generally when I encounter nicknames in a UI, I would not expect that changing the nickname would cause anything to break. However, by keying off of nickname, in fact we would break thing if someone changed a model nickname! We don't have a *great* way to do this currently: my first thought is that maybe we should key off of baseURL+(api)model name, but this actually won't work for OAuth models like Codex. Instead we should probably add a `generateModelId(...)` function that takes a model and generates some consistent ID for it: for non-codex models it would do `JSON.stringify({ apiBaseUrl, model })` and for codex models it would just do `JSON.stringify({ model, type: "codex" })`

reissbaker (changes requested):
Generally sensible approach but left a couple comments

source/session-history/schema/session-history-schema.ts
sdjidjev: For #1, you're totally right that history items all have access to a model. I was just thinking too far ahead in a world where for some reason octo did more than just "model stuff" for each history item (like workflows, etc. which wouldn't require a model necessarily). but in that case we can come up with a new way to do it; no need to mega-futureproof this!

Regarding #2, wouldn't we have to put a default value into `historyItem`s from the past that don't have a `model<X>` field? But regardless, if we're storing the field like #3 suggests then we can have the default value guaranteed to not overlap with anything else! My concern was only an issue if we were gonna use `modelNickname`.

I like your idea for #3!

reissbaker: Ohhhh I see, yes true for historical trees we'll need a default. Hrmm. I guess it's gotta be nullable on the DB side.

drizzle/0006_history_item_model_json.sql
```diff
@@ -0,0 +1,7 @@
+ALTER TABLE `history_items` ADD `model_json` text;
+--> statement-breakpoint
+UPDATE `history_items`
+SET `model_json` = 'octo-no-model-recorded'
+WHERE `model_json` IS NULL;
+--> statement-breakpoint
+ALTER TABLE `history_items` ALTER `model_json` SET NOT NULL;
```
sdjidjev, line 7: @reissbaker wondering if this is the right way to set up a non null modelJson field while still supporting the old `historyItem` rows that don't have a `modelJson` field. i'm worried that the backfill operation could be too slow or expensive. llm says it'll be super fast and nothing to worry about but thought i'd check in on that 😅 

alternatively we could do:
```
ALTER TABLE `history_items` ADD `model_json` text DEFAULT 'octo-no-model-
recorded' NOT NULL;
```
but then in typescript you could technically pass in a payload omitting the modelJson field and it'll accept that.

which option is better?

reissbaker: Backfill is cheap! Backfill sounds good to me :)

reissbaker (approved):
LGTM!

PR #252, opened 2026-07-28
Show toast when text is copied

sdjidjev:
Uses paintcannon's `clipboardWrite` event to display a toast when the user has copied text in octo.
Screenshot 2026-07-28 at 4 24 53 PM

Also adds special handling in `useToast` when the child is just a string (vs an element) since i noticed the formatting/borders were wonked without them

Before
Screenshot 2026-07-28 at 4 11 53 PM
After
Screenshot 2026-07-28 at 4 14 08 PM

However, with this change I noticed that `Span` doesn't preserve the colored version of the ⚠️ emoji. Clanker says paintcannon needs to operate on grapheme clusters instead of Rust `char`s. Gonna open up a issue in paintcannon for tracking purposes but the fix is not urgent!

### DIFF
--- a/source/app.tsx
+++ b/source/app.tsx
@@ -190,6 +190,7 @@ export default function App({
   bootSkills,
 }: Props) {
   const { paintCannon } = useApp();
+  const showToast = useToast();
   const [hasFocus, setHasFocus] = useState(paintCannon.hasFocus);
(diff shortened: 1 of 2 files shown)


fireph (approved):
🔥

billycao (approved):
neato burrito :)

PR #253, opened 2026-07-28
Add syn: model aliases instead of hard coded model strings

billycao:
Noticed even on latest octo the synthetic model imports are already out of date:

CleanShot 2026-07-28 at 16 09 48@2x

I know this will be fixed by libclank + reading our `/models` endpoint, but a small proposal for now to use the neat `syn:` aliases Matt put in for now, to avoid a poor experience. Context window might still drift, but at least this guides users to use our more durable endpoints for a better experience.

CleanShot 2026-07-28 at 16 32 10@2x


reissbaker (approved):
LGTM!

PR #254, opened 2026-07-29
Restore terminal title to original text on octo exit

sdjidjev:
Noticed we would set the title to `\_o_O.//` but then would leave it there even after octo was exited.

This PR restores it to what the title was before launch!


reissbaker (approved):
Oh nice! LGTM!

PR #255, opened 2026-07-29
Add volta node and npm versions

billycao:
Small proposal to pin node and npm versions with volta.

Volta is opt in but nice to document somewhere. Ran into something earlier where I was using npm 11.6 and there was an `package-lock.json` change generated by npm 11.8 that was checked in that my npm version was trying to revert :P


sdjidjev (approved):
didn't know about volta before; looks good to me!

PR #256, opened 2026-07-29
Fix tools possibly running multiple times when opening menu

reissbaker:
Since the current tool to run was stored in the ToolRequestsRenderer state, when we opened the menu (which unmounted the ToolRequestsRenderer) we lost track of which tools had already ran. If you opened the menu in the middle of a batch of tools running(i.e. if you had already run some of the tools, but not all of them), it would re-run the earlier tools when it was remounted on menu close. This generally would be bad, since e.g. edits might not re-apply, etc.

This fixes the bug by deriving the current tool to run from the Zustand state:

1. We always store the current tool ID (or null, if no tools are running) in Zustand
2. In the component, instead of storing state, we derive which tools haven't run yet by examining which tools don't have outputs yet by querying Zustand
3. We also skip the tool that's currently running and wait for it to finish , if one is running

This means that mounting or unmounting the component doesn't have any impact.

This change also includes some type-level changes to LlmIR in order to correctly determine which LlmIRs correspond to tool outputs and which don't (so that we can derive the unanswered tools from the current state). However, it does not change their actual in-memory shape, and so we don't need to do any IR migrations.

This bug was found+fixed by Kimi K3.

### DIFF
--- a/source/state.ts
+++ b/source/state.ts
@@ -173,8 +173,8 @@ export function answeredToolCallIds(history: readonly HistoryNode[]): Set<string
 }
 
 export type ToolAction =
-  | { kind: "wait"; req: ToolCallRequest }
-  | { kind: "pending"; req: ToolCallRequest }
+  | { kind: "in-flight"; req: ToolCallRequest }
+  | { kind: "ready"; req: ToolCallRequest }
   | { kind: "done" };
 
 /*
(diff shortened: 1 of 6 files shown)


source/state.ts
```diff
@@ -154,6 +162,45 @@ function appendAndPersistHistory(
...
+  for (const item of history) {
+    if (item.type !== "llm-ir") continue;
+    const id = answeredToolCallId(item.ir);
+    if (id != null) answered.add(id);
+  }
+  return answered;
+}
+
+export type ToolAction =
+  | { kind: "wait"; req: ToolCallRequest }
```
sdjidjev: to me: `ready`, `in-flight`, `done` (or something similar) is more clear to me what the different states refer to.

sdjidjev (approved):
LGTM just a naming idea~

also, should we also set `runningToolCallId: null` in `startNewSession` and `hydrateNewSession`?

source/state.ts
reissbaker: Done!

reissbaker:
Great point, added!

PR #257, opened 2026-07-30
Bump paintcannon version

reissbaker:
Bumps Paintcannon to 0.0.37, which fixes the tmux-with-focus-events-off issue where when you switch sessions in tmux and have focus events in tmux unconfigured, Paintcannon will fire a `blur` event, but not reliably fire a `focus` event when you switch back.

For reference: tmux has a bug in which zooming and unzooming a pane will sometimes drop either focus events or resize events, or both, or will drop nothing and will send them both. This is a bug in tmux, not in Paintcannon, but since tmux is popular, we have fallback code in Paintcannon that runs on resize when Paintcannon detects it's running in tmux, which will synthesize dropped events if necessary. However! If you do not have tmux configured to send focus events, which insanely is the default tmux setting, this means that on resize we may synthesize blur/focus events...  And switching sessions fires a resize event in tmux (so we will synthesize a `blur` event, since we check the tmux server and see you've switched active sessions and the current one is not active). We *also* technically will synthesize focus events *on resize*... but we do so by checking for whether the pane is currently active, and when switching back to a session with many panes, you may switch to a different pane in the same session, so while tmux may still fire a resize event, it might happen when the pane is not active, so Paintcannon won't synthesize the focus event. And since tmux by default does not send focus events, when you later switch panes to the Octo pane, Paintcannon won't get a focus event (since tmux didn't send one) and won't synthesize one (since it only synthesizes on resize, since it was intended to work around the zoom/unzoom bug in tmux).

In 0.0.37, if Paintcannon synthesizes a blur event, it will start polling tmux to check whether the pane has become active again. If it has, it will synthesize the focus event, rather than simply waiting for another resize before synthesizing.

This is horrendous but it is what it is. Browsers exist to paper over terrible platform issues like this, and Paintcannon is a browser-API-for-the-terminal.


sdjidjev (approved):
up & up!

PR #259, opened 2026-07-31
Automatically retry failed requests

billycao:
Noticed that I was getting `terminated` errors w/ Synthetic (presumably because not all of our jobs have graceful termination set up correctly as discussed), which would interrupt the agent flow when unchained mode was on.

Small proposal to add a retry for failed requests:

https://github.com/user-attachments/assets/698b6321-7473-422c-8523-923b8102af82

~~- Also added migration and first time setup for config so it's explicitly in the config.~~

Removed config re: feedback from matt.

Tested config migration and retries locally.


reissbaker (changes requested):
This is a good change, but it belongs in `trajectory-arc.ts` (similar to edit fixing). `trajectory-arc.ts` expects to take a set of LlmIR and keep running until the harness is required to take action; since this doesn't *require* the harness to take action, it should live in the trajectory arc. This will also unclutter `state.ts` somewhat. The shape I'd use is that `trajectoryArc` has a new callback type for retries, and the callback passes an abortController that can be used to abort the retry; then `state.ts` just needs a new `mode` that also stores the abort controller for the retry, and React can just read the mode, render it if it appears, and call `mode.abortController.abort()` if the user wants to stop retrying.

reissbaker (changes requested):
This PR has gotten messy. There's a bunch of unrelated code in here, e.g. session history changes, and overloading a single `abortController.abort` call to mean both "retry now" and also "abort" depending on which string gets passed in isn't super idiomatic JS. Let's sync offline about this.

billycao:
Going to just close this and re-approach from first principles and understanding the code instead of just telling a llm to do this blindly :P

PR #260, opened 2026-08-03
Allow queued messages while LLM is responding

sdjidjev:
The input box remains visible while a model is responding. After submitting a message, it queues up to be sent after the currently running tool call (or batch) is completed OR once the assistant is done sending a message. 
If multiples messages are queued up, they're coalesced as one message concatenated with "\n".  

Screenshot 2026-08-03 at 2 43 37 PM
Screenshot 2026-08-03 at 2 43 52 PM


source/state.ts
```diff
@@ -523,25 +578,33 @@ export const useAppStore = create<UiState>((set, get) => ({
...
     set({ query });
   },
 
+  enqueueUserMessage: msg => {
+    set(state => ({
+      queuedUserMessages: [...state.queuedUserMessages, { ...msg, id: nextQueuedMessageId++ }],
+    }));
+  },
+
+  _sendQueuedUserMessages: (session, config) => {
```
reissbaker: nit: worth calling this `_appendQueuedUserMessages`? Since this doesn't per se send any requests, it just moved them from the queue into the history

reissbaker (approved):
Tiny nit but LGTM!

PR #263, opened 2026-08-05
Cap tool outputs to be 20% of total context limit

sdjidjev:
If a tool's output is estimated to be >= 20% of the model's context limit, we give the model an error warning them the output is too high and to try again (but still give 200 characters preview so maybe the model has something to go off of).


reissbaker (approved):
LGTM!

PR #265, opened 2026-08-05
Lift attached images into the app store

billycao:
A small proposal to lift the image storage from the component local state (`useState()`) to the global zustand state, like the text prompt already is.

The motivation is #266, which does the "ESC when user hasn't output tokens yet rewinds state". To also restore images, we need them in the global state, since the rewind code can't otherwise access component local state.

---

### DIFF
--- a/source/state.ts
+++ b/source/state.ts
@@ -154,6 +154,7 @@ export type UiState = {
   byteCount: number;
   vimMode: "NORMAL" | "INSERT";
   query: string;
+  attachedImages: ImageInfo[];
   queuedUserMessages: readonly QueuedUserMessage[];
   readonly history: readonly HistoryNode[];
   clearNonce: number;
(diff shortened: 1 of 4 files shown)


source/state.ts
```diff
@@ -174,6 +175,7 @@ export type UiState = {
   setVimMode: (vimMode: "INSERT" | "NORMAL") => void;
   setModelOverride: (m: ModelConfig, session: Session) => void;
   setQuery: (query: string) => void;
+  setAttachedImages: (images: ImageInfo[] | ((prev: ImageInfo[]) => ImageInfo[])) => void;
```
sdjidjev: the union of array vs function isn't a common zustand pattern; i think it could be confusing for future callers to know when to use which type. 
it would be more idiomatic & clear to have `addAttachedImage` `removeLastAttachedImage` and `clearAttachedImages` in zustand

sdjidjev (commented):
definitely a fan of having attachedImages in zustand, but just a comment on the function to set it!

sdjidjev (approved):

source/state.ts
billycao: Great point, done! :) Excellent feedback.

billycao:
Ty for review! :)

PR #266, opened 2026-08-05
Rewind the prompt when ESC aborts an unanswered response

billycao:
Adds the feature we talked about where ESC before the model streams any non-reasoning tokens removes the prompt from history and restores it (text + images) to the input.

https://github.com/user-attachments/assets/8da179a4-d0d9-432d-820e-02a00fee1480

---

### DIFF
--- a/source/state.ts
+++ b/source/state.ts
@@ -227,17 +227,12 @@ function appendAndPersistHistory(
 }
 
 /*
- * Whether `prefix` is still a prefix of `full`, by node identity. An in-flight arc captures
- * a history copy at start; the arc's own handlers only ever append to the live history, so
- * at settle time the copy remains a prefix — unless something external (a prompt rewind, a
- * new session) changed history underneath the arc.
+ * better-sqlite3 surfaces foreign key violations as a plain Error whose message is
+ * "FOREIGN KEY constraint failed" — there's no error code to match on, so match the
+ * message. Appending history under a deleted tree node (e.g. a rewound prompt) hits this.
  */
-function isHistoryPrefix(prefix: readonly HistoryNode[], full: readonly HistoryNode[]): boolean {
-  if (prefix.length > full.length) return false;
-  for (let i = 0; i < prefix.length; i++) {
-    if (prefix[i].nodeId !== full[i].nodeId) return false;
-  }
-  return true;
+function isForeignKeyViolation(e: unknown): boolean {
+  return e instanceof Error && /FOREIGN KEY constraint failed/i.test(e.message);
 }
 
 /*
@@ -816,6 +811,12 @@ export const useAppStore = create<UiState>((set, get) => ({
   runAgent: async ({ config, transport, session }) => {
     get()._appendQueuedUserMessages(session, config);
     const historyCopy = [...get().history];
+    /*
+     * The tip of this arc's history chain. Emissions are chained under it (see onMessage)
+     * rather than the live history tip, so that if the arc's context is deleted mid-flight
+     * (a prompt rewind), SQLite rejects the arc's appends.
+     */
+    let arcHistoryTip = historyCopy.at(-1)?.nodeId ?? null;
     const abortController = new AbortController();
     let compactionByteCount = 0;
     let responseByteCount = 0;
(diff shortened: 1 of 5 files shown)


source/state.ts
```diff
@@ -910,9 +955,22 @@ export const useAppStore = create<UiState>((set, get) => ({
...
       });
       throttle.flush();
-      set({
-        history: appendAndPersistHistory(session, historyCopy, outputToHistory(finish.irs), model),
-      });
+      /*
+       * If the last prompt was rewound, don't persist the stale history item with the previous
+       * prompt and any stale in-flight response tokens.
+       */
+      if (get().promptRewindPending) {
```
reissbaker: This feels race-y to me, or at least unexpected in a way that could lead to future bugs. Basically it looks like:

1. Rewinding the prompt aborts the request and sets `promptRewindPending` to `true`
2. Various places set it to `false`, including below
3. If it's still `true` here, we avoid writing history, because we're probably a stale, aborted history tree that was rewound. We have to do this here because this will run after the request aborts.

It feels a little scary that we need to audit every place `promptRewindPending` is set in order to make this branch correct, and continue auditing it in the future, since setting it to `false` in a function that could get called prior to the code here running would cause bugs. Since the requests are all waiting on an `await`, they can technically run on the next VM microtick rather than immediately, so an intermediate function call could slip through before this runs.

I think a more robust design would be:

1. The state stores `rewoundUserMessageIds`, which is an empty `Set<string>`
2. When we rewind, we add the now-removed user message ID to `rewoundUserMessageIds`
3. When appending history, we walk our tree backwards to the most-recent user message and check whether it's in the rewound set. If it is, skip SQLite writes.

That being said, the *most* robust design is to delete the row in SQLite, and have this write throw an error due to the user ID parent being deleted, and us then catch the thrown error and move on: then there's no bookkeeping on our end, and SQLite ensures the schema is satisfied. Is that possible/easy with the current schema? My sense for the schema is that this is should be relatively simple to add support for. Currently, the tree node schema looks like:

```typescript
// ... a bunch of columns, and then:
parentId: integer(),
```
If we change this to: 
```typescript
// in tree nodes schema:
parentId: integer().references(() => treeNodes.id, {
  onDelete: "cascade",
})
```

I *think* should enforce that all parent IDs need to actually reference another ID in the table, and *should* throw an error when that isn't true. However, SQLite is loose enough that I'm not 100% confident (Postgres should enforce it, but SQLite's a little weird). Worth a shot trying though, I think this would simplify the PR a lot, and also get us free cleanup when subtrees get deleted.

reissbaker (changes requested):
Generally love this feature, but I think the way we're currently skipping writes is a little error-prone. Left two suggestions for how to improve it; I think the best option is at the SQLite-level, but I'm not 100% confident on SQLite enforcing the constraint; if it does and it works, that's great and we can simplify the PR a lot, otherwise I think the tombstone set of `rewoundUserMessageIds` is probably more robust.

source/state.ts
billycao: Great feedback.

Actually, the foreign key constraint is already there, later in the file:

```ts
    foreignKey({
      name: "tree_nodes_parent_same_tree_fk",
      columns: [table.parentId, table.treeId],
      foreignColumns: [table.id, table.treeId],
```

I had the clanker write `source/session-history/tree-node-fk.test.ts` to verify that sqlite properly enforces FK constraints as expected, and also additionally verified that the tests fail when `foreign_keys = OFF` in `source/db/db.ts`.

Instead of catching FK constraint errors, the clanker believes we can detect stale state with `isHistoryPrefix()`, and instead of throwing the error, it should propagate and be caught by `runAgent()` which will correctly leave the live history in the correct state.

I looked at the code and it seems reasonable, but I'm not as familiar with the code so defer to you. :)

PTAL!

billycao (commented):
Great feedback, PTAL! :)

reissbaker:
Great to know that SQLite is working :) 

In that case, I would disagree here with writing code in JS to do the same schema validation as SQLite is doing: I think if SQLite is enforcing the constraint, we should rely on the SQLite schema to do its job instead of defensively trying to also do it in JS (overly defensive "layers" of code doing the same thing is a bit of an LLM smell to me). We can catch and ignore the SQLite error, and we then no longer need the history prefix code.

billycao:
Sorry for delay :) @reissbaker PTAL.

Updated `source/state.tsx` to check for foreign key violation instead of checking manually with `isHistoryPrefix()`, and rebased on main. Only the latest commit contains any changes.

PR #268, opened 2026-08-11
Fix data duplication

reissbaker:
Previously, we had two bugs:

1. When appending history, we actually wrote quite a bit more into the DB than we expected. Since both `retryTool` and `compactionParsed` called `appendAndPersistHistory`, and then we called `appendAndPersistHistory` again on the returned IRs, we wrote some IRs to the DB twice! Thanks to our branch-handling code, this was actually fine; the earlier writes would be considered abandoned branches and we'd ignore them. But, still mildly annoying to be bloating disk space.
2. Sending an abort immediately after a compaction would drop the compaction IR, since we called `abort([])` instead of passing along the compaction IR.

This PR fixes both bugs, by redesigning `trajectoryArc` to be a bit easier to hold. Instead of returning the IRs at the end while also having some IR-specific handlers that run midway through (which you then need to track and not serialize at the end...), we strip it all out in favor of an `onMessage` handlers that runs for all IRs, and we no longer return the IRs at the end: if you want them, you need to use the `onMessage` handler.

### DIFF
--- a/source/agent/trajectory-arc.ts
+++ b/source/agent/trajectory-arc.ts
@@ -1,4 +1,5 @@
 import fs from "fs/promises";
+import { registry } from "antipattern";
 import type { OctoIR } from "../ir/octo-ir.ts";
 import type {
   Content,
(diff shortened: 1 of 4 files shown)


sdjidjev (approved):
great catch and great fix!

PR #269, opened 2026-08-12
Fix orphaned tool output

reissbaker:
This fixes the bug I reported in Slack the other day — it was possible to drop tool calls in two ways, which would break Kimi K3 since we wouldn't drop the resulting tool output (and Kimi always needs paired tool calls with tool output):

- We dropped mangled tool calls, since backends may not accept e.g. invalid JSON. To handle this pairing issue, we now send the response to the mangled tool calls as a `user` message in the standard compiler.
- If you aborted an autofix request, we accidentally dropped the assistant message that you aborted. To handle this we... don't do that :laughing: 

This PR also adds fix-on-hydrate code to automatically fix broken historical sessions that encountered this bug, since this would perma-break Kimi K3 sessions. To avoid silently covering up these kinds of errors in the future, in canary builds we loudly error on hydrate if there are any unpaired tool calls.

### DIFF
--- a/source/libocto/compilers/standard.ts
+++ b/source/libocto/compilers/standard.ts
@@ -137,7 +137,7 @@ ${JSON.stringify(requestBody)}
 JSON`;
 }
 
-async function toLlmMessages<A extends Agent<any, any, any>>(
+export async function toLlmMessages<A extends Agent<any, any, any>>(
   messages: Array<CompilerIR<A>>,
   systemPrompt?: () => Promise<string>,
   modalities?: CompilerModalities,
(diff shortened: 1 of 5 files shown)


source/libocto/compilers/standard.ts
```diff
@@ -213,10 +219,16 @@ function llmFromIr<A extends Agent<any, any, any>>(
     };
   }
 
+  /*
+   * A tool-parse-error's malformed call is not re-emitted as an assistant tool_call (see the
+   * assistant branch above), so there is nothing for a tool message's tool_call_id to resolve
+   * to — and strict backends hard-reject tool messages without a matching preceding tool_call.
+   * Deliver the failure as a plain user message instead; a user message is always legal, and
+   * the model can see its call malformed and recover.
+   */
```
sdjidjev, line 228: would it be helpful to the LLM  if we included the `toolCallId` in the `content.text` field as well?

sdjidjev (approved):
neat solution! glad it works!
just one comment/question i had!

source/libocto/compilers/standard.ts
reissbaker: In this case, no — since this is happening only in the malformed-request case, and in the malformed-request case we don't pass the original tool call back to the backend (since it's malformed, and some backends will reject it), so the LLM can't see what the original ID was.

PR #270, opened 2026-08-13
OctoProcessManager to spawn, manage, and kill child processes

sdjidjev:
To prepare for the background process tool in octo, making an OctoProcessManager to keep track of and clean up processes (on Octo exit).

~TODO~
~- [x] add a PR stacked ontop of this with eslint to ensure that you can't import from `"child_process"` directly (only use OctoProcessManager)~

### DIFF
--- a/source/cli/cli.tsx
+++ b/source/cli/cli.tsx
@@ -44,7 +44,7 @@ import { render, type CreateRootOptions } from "paintcannon-react";
 import { ToastProvider } from "../components/toast.tsx";
 import { setOctoTitles } from "./titles.ts";
 import changelog from "../../CHANGELOG.md" with { type: "text" };
-import { installGlobalProcessSignalHandlers, registerCleanup } from "../octo-process.ts";
+import { processes } from "../octo-process.ts";
 
 const INTERACTIVE_RENDER_OPTIONS = {
   alternateScreen: true,
@@ -71,7 +71,7 @@ function renderInteractive(element: React.ReactNode, options: { captureCtrlC: bo
 const CONFIG_STANDARD_DIR = path.join(os.homedir(), ".config/octofriend/");
 const CONFIG_JSON5_FILE = path.join(CONFIG_STANDARD_DIR, "octofriend.json5");
 
-installGlobalProcessSignalHandlers();
+processes.manager().installGlobalProcessSignalHandlers();
 
 const cli = withOctoFlags(
   new Command().description("If run with no subcommands, runs Octo interactively."),
@@ -242,7 +242,7 @@ async function runMain(opts: {
     restoreTitles();
     await opts.transport.close();
   };
-  const markCleanupAsComplete = registerCleanup(cleanup);
+  const unregisterCleanup = processes.manager().registerCleanup(cleanup);
 
   try {
     let { config, configPath } = await loadConfig(opts.parsedCliArgs.config);
@@ -310,7 +310,7 @@ async function runMain(opts: {
       }
     }
   } finally {
-    markCleanupAsComplete();
+    unregisterCleanup();
     await cleanup();
   }
 }
--- a/source/cli/cli.tsx
+++ b/source/cli/cli.tsx
@@ -237,10 +237,13 @@ async function runMain(opts: {
   const cleanup = async () => {
     if (cleanedUp) return;
     cleanedUp = true;
-    await shutdownLspClients();
-    await shutdownMcpClients();
     restoreTitles();
-    await opts.transport.close();
+    await Promise.all([
+      shutdownLspClients(),
+      shutdownMcpClients(),
+      opts.transport.close(),
+      processes.manager().terminateAll(),
+    ]);
   };
   const unregisterCleanup = processes.manager().registerCleanup(cleanup);
 
--- a/source/cli/cli.tsx
+++ b/source/cli/cli.tsx
@@ -44,6 +44,7 @@ import { render, type CreateRootOptions } from "paintcannon-react";
 import { ToastProvider } from "../components/toast.tsx";
 import { setOctoTitles } from "./titles.ts";
 import changelog from "../../CHANGELOG.md" with { type: "text" };
+import { installGlobalProcessSignalHandlers, registerCleanup } from "../octo-process.ts";
 
 const INTERACTIVE_RENDER_OPTIONS = {
   alternateScreen: true,
@@ -69,6 +70,8 @@ function renderInteractive(element: React.ReactNode, options: { captureCtrlC: bo
 
 const CONFIG_STANDARD_DIR = path.join(os.homedir(), ".config/octofriend/");
 const CONFIG_JSON5_FILE = path.join(CONFIG_STANDARD_DIR, "octofriend.json5");
+
+installGlobalProcessSignalHandlers();
 
 const cli = withOctoFlags(
   new Command().description("If run with no subcommands, runs Octo interactively."),
@@ -230,6 +233,17 @@ async function runMain(opts: {
     session = useAppStore.getState().startNewSession(opts.transport.cwd, opts.parsedCliArgs);
   }
 
+  let cleanedUp = false;
+  const cleanup = async () => {
+    if (cleanedUp) return;
+    cleanedUp = true;
+    await shutdownLspClients();
+    await shutdownMcpClients();
+    restoreTitles();
+    await opts.transport.close();
+  };
+  const markCleanupAsComplete = registerCleanup(cleanup);
+
   try {
     let { config, configPath } = await loadConfig(opts.parsedCliArgs.config);
 
@@ -296,10 +310,8 @@ async function runMain(opts: {
       }
     }
   } finally {
-    await shutdownLspClients();
-    await shutdownMcpClients();
-    restoreTitles();
-    await opts.transport.close();
+    markCleanupAsComplete();
+    await cleanup();
   }
 }
 
--- a/source/lsp/client.ts
+++ b/source/lsp/client.ts
@@ -11,7 +11,7 @@ import type {
   Range,
 } from "vscode-languageserver-types";
 import { DiagnosticSeverity, SymbolKind } from "vscode-languageserver-types";
-import { type OctoProcess, OctoProcessManager } from "../octo-process.ts";
+import { type OctoProcess, processes } from "../octo-process.ts";
 
 // https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#headerPart
 const HEADER_DELIMITER_STRING = "\r\n\r\n";
@@ -118,7 +118,7 @@ const CLIENT_CAPABILITIES = {
 };
 
 export class LspClient {
-  private readonly octoProcessManager = new OctoProcessManager();
+  private readonly octoProcessManager = processes.manager();
   private octoProcess: OctoProcess | null = null;
   private requestIdCounter = 1;
   private pendingRequests = new Map<number, PendingRequest>(); // indexed by requestId
--- a/source/lsp/client.ts
+++ b/source/lsp/client.ts
@@ -1,4 +1,3 @@
-import { spawn, type ChildProcess } from "child_process";
 import { type Writable } from "node:stream";
 import path from "path";
 import type {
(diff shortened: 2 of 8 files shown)


source/lsp/client.ts
```diff
@@ -136,16 +137,17 @@ export class LspClient {
...
     });
+    const childProcess = this.octoProcess.childProcess;
 
-    const stdout = this.process.stdout;
+    const stdout = childProcess.stdout;
     if (!stdout) throw new Error(`LSP server "${this.serverConfig.serverName}" has no stdout`);
     stdout.on("data", (chunk: Buffer) => this.onData(chunk));
 
-    this.process.on("error", err => {
+    childProcess.on("error", err => {
```
fireph: crazy idea, what if octoProcess had this same interface? (`on`: `error`, `exit`, etc.) is that a bad idea? that way you wouldn't have to do this `const childProcess = this.octoProcess.childProcess;` and keep the reference around to use the octo process api.

source/lsp/client.ts
```diff
@@ -335,35 +339,36 @@ export class LspClient {
     timeoutMs: number = REQUEST_TIMEOUT_MS,
   ): Promise<any> {
     return new Promise((resolve, reject) => {
-      const stdin = this.process?.stdin;
+      const stdin = this.octoProcess?.childProcess.stdin;
```
fireph: same here, wonder if there would be a nice way to have octoProcess inherit the api of childprocess without a ton of extra code. if kimi k3 says this idea is stupid, feel free to ignore tho

source/lsp/client.ts
```diff
@@ -335,35 +339,36 @@ export class LspClient {
...
       if (!stdin?.writable) {
         reject(new Error(`LSP server "${this.serverConfig.serverName}" is not running`));
         return;
       }
 
       const requestId = this.requestIdCounter++;
+      const timer = setTimeout(() => {
```
fireph: this seems like it is fixing a different thing than child process stuff. should this be a separate PR just to fix request timeouts?

source/octo-process.ts
```diff
@@ -0,0 +1,272 @@
...
+  execFile,
+  spawn,
+  type ChildProcess,
+  type ExecFileException,
+  type ExecFileOptions,
+  type SpawnOptions,
+} from "child_process";
+
+/**
+ * Child process lifecycle management, over `child_process`.
```
fireph: `over` maybe could be reworded to `containing` or something like that.

source/octo-process.ts
```diff
@@ -0,0 +1,272 @@
...
+ *
+ * All spawning goes through {@link OctoProcessManager}; every process it
+ * spawns carries a `terminate` callback and is terminated when Octo exits
+ * (unless spawned with `surviveExit`). {@link registerCleanup} and
+ * {@link installProcessSignalHandlers} run shutdown logic even on
+ * SIGINT/SIGTERM/SIGHUP, where `try/finally` blocks never run.
+ */
+
+type OctoProcessOptions = {
+  detached?: boolean;
```
fireph, line 31: maybe document what `detached` means here

source/octo-process.ts
```diff
@@ -0,0 +1,272 @@
...
+export type OctoExecFileOptions = ExecFileOptions & OctoProcessOptions;
+
+/** `stdout`/`stderr` are strings unless `encoding: "buffer"` is passed, mirroring `child_process.execFile`. */
+export type OctoExecFileCallback = (
+  error: ExecFileException | null,
+  stdout: string | Buffer,
+  stderr: string | Buffer,
+) => void;
+
+// Processes terminated when Octo exits.
```
fireph: useless comment imo

source/octo-process.ts
```diff
@@ -0,0 +1,272 @@
...
+  handler: () => void,
+): Promise<void> {
+  if (handlingTerminationSignal) return;
+  handlingTerminationSignal = true;
+
+  try {
+    await runCleanups();
+  } finally {
+    // Re-raise without our handler so the process dies from the signal
+    // itself, exiting with the conventional 128 + signo code.
```
fireph: is `signo code` correct? didn't use that terminology above

source/state.ts
```diff
@@ -339,6 +335,7 @@ export const useAppStore = create<UiState>((set, get) => ({
     const timer = setTimeout(async () => {
       await runNotifyCommand(config);
     }, notifyTimeout);
+    timer.unref?.();
```
fireph: is this worth documenting?

source/lsp/client.ts
sdjidjev: oh yeah for sure, i originally did this PR to stop runaway processes but it's def better to separate the fixes out.

source/state.ts
sdjidjev: this is good to put in the separate PR with that other timeout change

source/octo-process.ts
```diff
@@ -0,0 +1,267 @@
...
+  /** signal the whole process group on terminate */
+  declare detached?: boolean;
+  /** don't kill when Octo exits */
+  declare surviveAfterOctoExit?: boolean;
+
+  constructor(childProcess: ChildProcess, options: OctoProcessOptions) {
+    super();
+    // Since Node creates processes via factories (spawn), we can't use `extends` directly
+    const octoProcess = childProcess as OctoProcess;
+    Object.setPrototypeOf(octoProcess, OctoProcess.prototype);
```
reissbaker: In browser-JS-land it's generally considered an antipattern to extend native/host classes. While extending `ChildProcess` in Node *works*, I think this is a somewhat unintuitive interface and requires doing non-TypeScript-y things such as manually overriding the prototype chain here, and using `declare` to claim variables.

Rather than trying to 1:1 mimic the Node process interface, I think it would be simpler to just write a wrapper class with the exact exposed functions we actually use in Octo, and not modify the prototype chain of host objects; instead, just store the `childProcess: ChildProcess` as a private variable in the wrapper `OctoProcess` and call functions on it as necessary.

We already mostly don't call `child_process` manually — from a quick ripgrep it looks like we only use it in `config.ts`, `browser-spawn.ts`, `lsp/client.ts`, and hidden by various transports with a narrow API. We can probably expose a narrow API here too.

A quick read makes me think a typed EventEmitter with [`typed-emitter`](https://github.com/andywer/typed-emitter) would probably suffice: just copy the `stdout`, `stderr`, `stdin` pipes to the wrapper object, expose `kill` and `terminate` functions, and emit the `error` and `exit` events.

source/octo-process.ts
```diff
@@ -0,0 +1,267 @@
...
+/** `stdout`/`stderr` are strings unless `encoding: "buffer"` is passed, mirroring `child_process.execFile`. */
+export type OctoExecFileCallback = (
+  error: ExecFileException | null,
+  stdout: string | Buffer,
+  stderr: string | Buffer,
+) => void;
+
+const octoProcesses = new Set<OctoProcess>();
+
+const cleanups = new Set<() => Promise<void> | void>();
```
reissbaker: It's somewhat surprising to me that all process manager instances share a global set of processes and cleanups — and global variables also make tests more likely to leak internal state across each other. IMO the set of processes and cleanup functions should be owned by the class, rather all instantiations sharing global state.

My guess is the reason you wrote it this way was because passing around a single instance of OctoProcessManager all over the place is annoying, and if all instances share global state, it's less annoying. But instead of global state, you can solve that problem using our dependency injection framework [`antipattern`](https://github.com/reissbaker/antipattern):

```typescript
import { registry } from "antipattern";

const manager = new OctoProcessManager();
export const process = registry({
  manager: () => {
    return manager;
  },
});
```

Later, in calling code, you can do:

```typescript
// instead of const manager = new OctoProcessManager(...);
const manager = process.manager();
```

And in tests, mocking out all process management is easy:

```typescript
const mockedProcessManager = {
  spawn: vi.fn(),
  // ...
};
withMock(process, "manager",  () => mockedProcessManager, () => {
  // all caller attempts to access process.manager will
  // instead hit mockedProcessManager

  // so you can call a bunch of code, and then:
  expect(mockedProcessManager.spawn).toHaveBeenCalled();
});
```

While this still leaves a little bit of shared global state — the `manager` instance — it makes it relatively clean to mock out during tests, and not accidentally leak parts of the class you might not expect to be shared. Having lots of internal global state means anyone writing tests will have to be aware of (and stay up to date with) the implementation of `OctoProcessManager`, since you need to mock out *all* of the module's global variables (and add more mocking if more globals are added); storing the state in the class and either passing it around or using dependency injection makes it easier to maintain tests long term.

source/octo-process.ts
```diff
@@ -0,0 +1,267 @@
...
+
+const SIGKILL_GRACE_MS = 1000;
+const CLEANUP_TIMEOUT_MS = 5000;
+const FORCE_EXIT_FALLBACK_MS = 1000;
+
+const TERMINATION_SIGNALS = ["SIGINT", "SIGTERM", "SIGHUP"] as const;
+type TerminationSignal = (typeof TERMINATION_SIGNALS)[number];
+
+// Conventional exit codes: 128 + signal number
+const SIGNAL_EXIT_CODES: Record<TerminationSignal, number> = {
```
reissbaker, line 121: Technically I believe `Record<TerminationSignal, number>` isn't required to be exhaustive. To force exhaustiveness checks (valuable IMO!) you could rewrite this as:

```typescript
const SIGNAL_EXIT_CODES: { [K in TerminationSignal]: number } = {
  // ...
};
```

Could be wrong though, but I suspect deleting e.g. line 82 wouldn't cause the compiler to fail here, whereas with the type above it would.

source/octo-process.ts
```diff
@@ -0,0 +1,267 @@
...
+    }
+    killAllOctoProcesses();
+  })();
+
+  let timer: NodeJS.Timeout | undefined;
+  try {
+    await Promise.race([
+      work,
+      new Promise<void>(resolve => {
+        timer = setTimeout(resolve, CLEANUP_TIMEOUT_MS);
```
reissbaker: We have a helper function for this :)

```typescript
import { sleep } from "./sleep.ts";

await Promise.race([
  work,
  sleep(CLEANUP_TIMEOUT_MS)
]);
```

reissbaker (changes requested):
Generally good approach, but left some feedback on some cleanup opportunities!

source/octo-process.ts
sdjidjev: checked by deleting and it is exhaustive as-is!

sdjidjev:
thanks for the feedback!! ready for another review @reissbaker

source/transports/local.ts
```diff
@@ -9,13 +8,19 @@ import {
...
 } from "./transport-common.ts";
+import { processes } from "../octo-process.ts";
+
+const KILL_GRACE_MS = 500;
 
 const STRIPPED_ENV_VARS = ["NODE_ENV", "NAPI_RS_NATIVE_LIBRARY_PATH", "CANARY_OCTO"];
 
 export class LocalTransport implements Transport {
   cwd = process.cwd();
+  private readonly octoProcessManager = processes.manager();
```
reissbaker: Nit: IMO `octoProcessManager` shouldn't use the shared global process manager; instead it should instantiate its own via `new OctoProcessManager()`. In `close()`, we call `this.octoProcessManager.terminateAll(...)`, and since we're using the shared global process manager, that terminates everyone's processes — not just the processes spawned by this instance of `LocalTransport`.

In practice, I don't think we currently ever close a `LocalTransport` without *also* intending to shut down Octo entirely and kill all processes, but that just happens to be true; in theory we could have multiple local transports, or swap transports midway through, etc. In that case we wouldn't want to kill all processes, only the ones associated with this particular transport.

Since `OctoProcessManager` is now able to be created without sharing global state, I think it's worth instantiating a new one for the LocalTransport; there's no harm (since we call `.close` on it anyway), and it might help understand what's going on in the future.

source/octo-process.ts
reissbaker: Sounds good!

reissbaker (approved):
Left one nit but LGTM!

source/cli/cli.tsx
```diff
@@ -230,6 +233,17 @@ async function runMain(opts: {
...
   }
 
+  let cleanedUp = false;
+  const cleanup = async () => {
+    if (cleanedUp) return;
+    cleanedUp = true;
+    await shutdownLspClients();
+    await shutdownMcpClients();
+    restoreTitles();
+    await opts.transport.close();
```
reissbaker: Assuming we make `transport.close()` only shut down its own processes, I think you'll want to add `process.manager().terminateAll(...)` here as well to shut down any global processes spawned.

reissbaker: Also minor nit — since these all have some degree of graceful term, and we've got a bunch of these cleanups now, it might be worth doing something like:

```typescript
restoreTitles();
await Promise.all([
  shutdownLspClients(),
  shutdownMcpClients(),
  opts.transport.close(),
  processes.manager().terminateAll(...);
]);
```

So that these run in parallel and we only wait `MAX(termTime)` instead of `SUM(termTime)`

PR #271, opened 2026-08-14
OctoBun

reissbaker:
Ports Octo to Bun, and gets single-file-executable compilation working, including cross-platform builds for Linux x64, ARM, and old pre-AVX2-x64, as well as the same for Darwin (aka macOS).

This means we don't need our users to monkey around with node versions, npm, etc; we just compile an `octo` executable, and it runs, using exactly the Bun version and packages we precompiled in. We port the package's `bin` files to a Bash script that autodetects the user's platform and launches the correct precompiled executable. In the future, we could also support non-npm-based installs; for example, a curl-to-bash that just stuffs the right executable somewhere and adds it to the `$PATH`, for easy auto-update without needing permissions or unsafe `chown`s; Claude Code stuffs theirs into `~/.claude`, which is obnoxious; we should probably use `~/.local/bin/octofriend` and a symlink from `~/.local/bin/octo` to the latter.

The Bun port also fixes (in a sense) the `better-sqlite3` segfault on Node 22, since we no longer use `better-sqlite3` (Bun has a native SQLite library built-in that Drizzle knows how to use), and we also no longer use Node.

Note that this sets `tsc` to `noEmit`, since Bun can handle .ts and .tsx files natively. Currently we still need to use Babel+a Bun plugin in order to run the React Compiler; in Bun 1.4 it's natively integrated into Bun and so once that's released we should upgrade and opt into using it, and delete Babel and the plugin. I tried leaving out the React compiler, but on very long conversations, typing started lagging; putting the compiler back in fixed the perf issues.

There were a few compatibility issues that needed to be fixed as part of this port:

1. Bun-on-macOS uses macOS's own SQLite library, which is pre-3.5, and thus doesn't support all of the syntax we used for some handwritten migrations (Drizzle-generated migrations work correctly, but we had an alter column migration that requires SQLite 3.5+ that appears to have been handwritten, or at least manually created by an LLM rather than by `drizzle-kit generate`).
2. Vitest does not understand some Bun-isms, especially importing static files (required for single-file-executables)
3. Bun 1.3 can't embed directories into static executables, only files, which caused issues for Drizzle's migration code which expects directories of migration files.

I had Kimi K3 rewrite the one migration file that used the 3.5+ SQLite code to fix issue (1). For (2) I had Kimi port our test suite from Vitest to [Bun's native test runner](https://bun.com/docs/test), which has the side benefit of improving test speeds. This was largely mechanical, since Bun's native test runner is mostly compatible with Vitest (since both are designed based on Jest), although when it encountered Vite-esque mocking I directed it to port it to the [antipattern](https://github.com/reissbaker/antipattern) library we use for mocking, which works regardless of test suite platform. Note that there are still a few places where we call `bun:test`'s `mock` function; that's because Bun's `mock` is also how you create a spy, and so we use it to spy on function calls we pass in.

For directory embedding... Bun 1.4 supports a new `--asset` flag to compile entire directories into the executable, which work with `fs` functions like listing directories, reading files, etc. However, Bun 1.4 is not released yet; I tested it via Bun's canary build and it works great, but this feature doesn't exist yet in stable Bun. Instead, you need to import the files in your code, and *file reads* work but directory operations don't. Unfortunately, Drizzle's migrator expects to be able to read the directories.

As a temporary Bun 1.3 workaround, we use compile-time code gen in our new `build.ts` file to generate imports for every Drizzle migration, and then on Octo boot we write them all to a temp dir, and pass that temp dir to Drizzle's migrator.

Once Bun 1.4 is released we should upgrade and use `--asset` instead! This will also speed up boot time by some number of milliseconds, since it'll use a virtual filesystem in-memory instead of needing to write all the files and then read them back from the real fs.


sdjidjev:
1Dpi5G71

PR #273, opened 2026-08-14
Add ESLint & restrict imports from "child_process"

sdjidjev:
Added eslint on octofriend to use `no-restricted-imports` from `"child_process"`

Also adds `bun run lint` step to CI.

All child processes must be spawned via OctoProcess (source/octo-process.ts) so they're tracked and killed when Octo exits. ESLint is configured with the typescript-eslint recommended ruleset, with deliberately-used patterns (`any`, @ts-ignore) disabled and ML training tooling exempted from the child_process restriction since it runs standalone.

A few exceptions in the `eslint.config.js` file:
- `no-explicit-any` would need its own PR to fix rule-breakers for this
- `no-empty-object-type` - usecase of this in libocto, probably can keep.
- `no-empty` → `{ allowEmptyCatch: true }`
  - some empty catch blocks make sense for cleanups
- certain unused vars are allowed
  - `_`-prefixed
  - unused `catch(e)` binding
  - `const { skipMe, ...rest } = obj` it's ok if `skipMe` is unused.

---

### DIFF
--- a/source/components/file-suggestions/use-file-search.ts
+++ b/source/components/file-suggestions/use-file-search.ts
@@ -8,39 +8,6 @@ const MAX_SUGGESTIONS = 8;
 const CACHE_TTL = 5000;
 const GITIGNORE_CACHE_TTL = 300000;
 
-const FOLDERS = new Set([
-  "node_modules",
-  "bower_components",
-  ".pnpm-store",
-  "vendor",
-  ".npm",
-  "dist",
-  "build",
-  "out",
-  ".next",
-  "target",
-  "bin",
-  "obj",
-  ".git",
-  ".svn",
-  ".hg",
-  ".vscode",
-  ".idea",
-  ".turbo",
-  ".output",
-  "desktop",
-  ".sst",
-  ".cache",
-  ".webkit-cache",
-  "__pycache__",
-  ".pytest_cache",
-  "mypy_cache",
-  ".history",
-  ".gradle",
-]);
-
-const FOLDER_PATTERNS = [...FOLDERS].map(f => `${f}/**`);
-
 type FileListEntry = {
   files: string[];
   timestamp: number;
--- a/source/libocto/result.ts
+++ b/source/libocto/result.ts
@@ -102,7 +102,7 @@ function isResult(value: unknown): value is Result<any, any> {
 }
 
 export class Ok<T> implements IResult<T, any> {
-  success = true as const;
+  success: true = true;
   constructor(readonly data: T) {}
 
   map<New>(fn: (t: T) => New) {
@@ -127,7 +127,7 @@ export class Ok<T> implements IResult<T, any> {
 }
 
 export class Err<E> implements IResult<any, E> {
-  success = false as const;
+  success: false = false;
   constructor(readonly error: E) {}
 
   map<New>(_: (t: any) => New) {
--- a/source/libocto/result.ts
+++ b/source/libocto/result.ts
@@ -102,7 +102,7 @@ function isResult(value: unknown): value is Result<any, any> {
 }
 
 export class Ok<T> implements IResult<T, any> {
-  success: true = true;
+  success = true as const;
   constructor(readonly data: T) {}
 
   map<New>(fn: (t: T) => New) {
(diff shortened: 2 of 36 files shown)


source/components/file-suggestions/use-file-search.ts
```diff
@@ -8,39 +8,6 @@ const MAX_SUGGESTIONS = 8;
...
-  ".cache",
-  ".webkit-cache",
-  "__pycache__",
-  ".pytest_cache",
-  "mypy_cache",
-  ".history",
-  ".gradle",
-]);
-
-const FOLDER_PATTERNS = [...FOLDERS].map(f => `${f}/**`);
```
reissbaker, line 42: It looks like FOLDER_PATTERNS was accidentally unused, and this deleted it. Probably instead this should be fixed (use the folder patterns to ignore searching these common patterns)

source/libocto/result.ts
```diff
@@ -127,7 +127,7 @@ export class Ok<T> implements IResult<T, any> {
 }
 
 export class Err<E> implements IResult<any, E> {
-  success: false = false;
```
reissbaker, line 130: Dislike this. Can we turn off whatever lint this is? We should be able to specify types, IMO much more readable.

source/markdown.test.ts
```diff
@@ -286,6 +286,7 @@ function renderToString(markdown: string): string {
 }
 
 function stripAnsi(str: string): string {
+  // eslint-disable-next-line no-control-regex -- intentionally strips ANSI escape sequences
```
reissbaker: Can we disable this lint rule? IDK why we wouldn't allow regex control characters, they're useful.

training/fix-json/generate-broken-json.ts
```diff
@@ -349,7 +349,7 @@ function stringifyNode(node: JSONASTNode, outputNodes: OutputNode[]): string {
     case "boolean":
       return JSON.stringify(node.value);
 
-    case "array":
+    case "array": {
```
reissbaker: I think indentation is sufficient, which Prettier enforces. Can we disable this lint rule?

reissbaker (changes requested):
I think this is too large of a lint change, and after scrolling through a bunch of changes it felt like there was a lot of linting I'd prefer to turn off.

Rather than turning on all the "recommended" lints, can we instead just turn on the specific lint rules we want? It seems like the main goal is to prevent importing `child_process`. The no unused variables and no unused import lint rules also seemed useful, since Prettier apparently can't autofix those; maybe we just enable those two + our custom one banning `child_process`?

Also, `bun run lint` should run both ESLint and Prettier, since we use both linters. And `format` should also call ESLint, and we should probably remove `format:check` since that's just running Prettier as a linter (which we should add to `lint`).

source/components/file-suggestions/use-file-search.ts
sdjidjev: ahh ok, i had assumed we removed the folder checks intentionally but forgot to delete this.
i can add this back in another PR!

sdjidjev: looking through, looks like we can safely delete these since the excluded directories are recorded in [`transport-common.ts`](https://github.com/synthetic-lab/octofriend/blob/main/source/transports/transport-common.ts#L61) already!

source/markdown.test.ts
sdjidjev: til from kimi this is usually enabled to avoid weird invisible character typo bugs 🤔 

> ```
> // Looks like:  /^\d{3}-\d{4}$/
> // Actually is: /^\d{3}-\x07\d{4}$/   (invisible BEL char)
> ```
> The regex renders as a normal phone-number validator in your editor, but it now requires a BELL character between the two digit groups. Every real phone number fails validation, and because the source looks right, you can stare at it for an hour without spotting it.

sdjidjev (commented):
@reissbaker thanks for your comments, asked kimi to do a thorough review of what rules are either 1. covered by Prettier, or 2: genuinely bug-catching rules (as opposed to purely stylistic).
the only slight exception to this is `prefer-const` for non-reassigned values which i think will just help us keep the codebase cleaner.

PTAL!

reissbaker (approved):
LGTM!

PR #276, opened 2026-08-19
Autofix more tool pairing bugs

reissbaker:
I found a couple more tool pairing bugs thanks to our canary builds now throwing errors on tool pairing issues. While Octo *no longer generates these*, old sessions might still have them stored in the DB, so it's worth autofixing. This autofixes the ones I found and adds tests.


sdjidjev (approved):
thanks for the fix!

PR #277, opened 2026-08-19
Memoize input

reissbaker:
Tiny perf improvement, although I think it does help typing latency — uses `React.memo` to prevent re-rendering the input box unless the props change. We used to do this, but an Ink upgrade broke it (due to a React bug! `useEffectEvent` is known-broken when used inside React.memo, and Ink swapped their text input to be based on `useEffectEvent`).

Since Paintcannon doesn't rely on `useEffectEvent`, we can memoize this again.


sdjidjev (approved):
memos are so back 😎

PR #278, opened 2026-08-20
Update to Bun 1.4

reissbaker:
Updates Octo to use Bun 1.4, since it was just released. This allows us to embed entire directories into Octo, meaning we don't need the codegen-drizzle-imports song and dance and can just embed the entire drizzle dir.

It also seems to reduce memory usage somewhat for long conversations.

In theory this also should have allowed us to use Bun's built-in React Compiler for builds, since that shipped alongside the 1.4 release, instead of the Babel-based plugin we currently use. However, their React Compiler appears to be broken and crashes on boot, so we can't use it and thus must continue using our existing plugin setup. I'll report the crash upstream and hopefully in a later version of Bun they'll fix it, and we can stop using the Babel plugin and instead use the native Rust-based compiler in Bun.


sdjidjev (approved):
🎉

PR #280, opened 2026-08-26
Add background-process and manage-background-process tools

sdjidjev:
New tool pair for long-running shell commands:
- **background-process** spawns a command in its own process group (as an OctoProcess, so automatically terminates on Octo's exit)
- **manage-background-process**: has three actions [poll, kill, list] **polls** and **kills** existing running background processes, **lists** all running background process

Decided to make this as a pair of tools rather than coalesced into one in order to manage the permissions better.
The **background-process** tool requires permission whereas **manage-background-process** is always allowed to run.

### DIFF
--- a/source/background-process.ts
+++ b/source/background-process.ts
@@ -23,6 +23,7 @@ export class BackgroundProcess {
   private _outputExceeded = false;
   private _status: BackgroundProcessStatus = { state: "running" };
   private readonly activityListeners = new Set<() => void>();
+  private readonly processClosedPromise: Promise<void>;
 
   constructor(id: string, octoProcess: OctoProcess, command: string) {
     if (octoProcess.stdout == null || octoProcess.stderr == null) {
@@ -31,6 +32,7 @@ export class BackgroundProcess {
     this.id = id;
     this.octoProcess = octoProcess;
     this.command = command;
+    this.processClosedPromise = new Promise(resolve => octoProcess.once("close", () => resolve()));
 
     octoProcess.stdout.on("data", data => this.appendOutput(data));
     octoProcess.stderr.on("data", data => this.appendOutput(data));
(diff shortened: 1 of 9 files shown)


source/background-process.ts
```diff
@@ -0,0 +1,152 @@
...
+      throw new Error("Background processes must be spawned with piped stdio");
+    }
+    this.id = id;
+    this.label = label;
+    this.octoProcess = octoProcess;
+    this.command = command;
+    this.processClosedPromise = new Promise(resolve => octoProcess.once("close", () => resolve()));
+
+    octoProcess.stdout.on("data", data => this.appendOutput(data));
+    octoProcess.stderr.on("data", data => this.appendOutput(data));
```
fireph: do we want the LLM to not be able to differentiate stdout from stderr? I wonder if they should stay separated?

source/background-process.ts
```diff
@@ -0,0 +1,152 @@
...
+    await this.processClosedPromise;
+  }
+
+  get hasUnreadOutput(): boolean {
+    const buffered = this.output.getOutput();
+    return buffered != null && buffered.length > this.readOffset;
+  }
+
+  drainUnreadOutput(): string {
+    const buffered = this.output.getOutput();
```
fireph: nit: getOutput() reads entire output rather than just the new output since last drain.

the clanker says that: `A chunk-cursor would make polls O(new output).`

but this is a relatively minor optimization

fireph (approved):
lgtm! my only thought is that making the AI come up with a unique name for the process seems slightly risky. maybe it just gives a name and then an ID is attached to the end so that it is always unique?

sdjidjev:
@fireph I introduced the `label` field more as a way to let the user know what kind of background process is being run right now. the AI never subsequently queries via `label`, only via `id`. Would it be more clear if I named it `humanReadableLabel` or something to better describe that it won't be used by AI?

EDIT: spoke offline and agreed just rewording the comment to make it clear it won't be used by the AI for reference and just to be human-understandable.

source/background-process.ts
sdjidjev: Good point, I've separated them!

source/background-process.ts
sdjidjev: ah interesting, yeah i agree! made it use a chunk counter rather than a cursor

sdjidjev:
@fireph thank you for your review! i've addressed comments, PTAL!

PR #281, opened 2026-08-27
Fix disappearing input during tool calls

sdjidjev:
A bug was reported that the input field disappears during tool calls when using the queue messages feature.

By adding a new mode `tool-call-request`, we can determine whether we show the input based on the `mode` octo is in.

Also fixed a long-standing issue of checking the permissions whitelist being async and potentially showing the wrong state for a frame with a quick fix (since it's related to the input field i added it to this PR)

### DIFF
--- a/source/app.tsx
+++ b/source/app.tsx
@@ -717,50 +717,12 @@ function BottomBarContent({ inputHistory }: { inputHistory: InputHistory }) {
     },
     [query, modeData.mode, config, transport, session, setQuery, showToast],
   );
-  if (modeData.mode === "error-recovery") return <Loading />;
-  if (modeData.mode === "payment-error") {
-    return <PaymentErrorScreen error={modeData.error} />;
-  }
-  if (modeData.mode === "rate-limit-error") {
-    return <RateLimitErrorScreen error={modeData.error} />;
-  }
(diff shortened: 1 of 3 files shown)


source/app.tsx
```diff
@@ -1579,13 +1576,18 @@ function ToolRequestRenderer({
...
+    runTool,
+    notifyReadyForInput,
+  ]);
   if (noConfirmationNeeded || isRunning) {
-    return (
-      <Loading
-        overrideStrings={["Waiting", "Watching", "Smiling", "Hungering", "Splashing", "Writhing"]}
-      />
-    );
+    return null;
```
sdjidjev, line 1590: no longer needed since it's covered by `BottomBarContent` now!

Image

source/state.ts
```diff
@@ -102,6 +102,11 @@ export type UiState = {
         toolReqs: ToolCallRequest[];
         abortController: AbortController;
       }
+    | {
+        mode: "tool-call-request";
```
fireph: request is a bit of a loaded term in octo already, maybe it could be `tool-call-approval` or `tool-call-permission` or something along those lines.

fireph (approved):
Image

PR #282, opened 2026-08-28
Vim mode is INSERT when input field is available

sdjidjev:
Once a message is sent in Vim mode, keep it in INSERT mode (now that we can queue messages). Because ESC previously interrupted an ongoing LLM response (even in Vim mode), we need a new instant control sequence to abort a response even in INSERT mode.

After some in-person bike shedding, folks agreed on **Ctrl+ESC** being that sequence. HOWEVER, this only works on kitty keyboard enabled terminals. For Mac's Terminal.app, we just change the hint to say `(Press ESC twice to interrupt)`.

Kitty Keyboard Terminal Behavior:

https://github.com/user-attachments/assets/9cd76c47-12a5-4592-8627-c349a0aa761a

Non-Kitty Keyboard Terminal Behavior:

https://github.com/user-attachments/assets/5abbcb05-5fb1-4720-a4b3-6abc19902730

### DIFF
--- a/source/components/text-input.tsx
+++ b/source/components/text-input.tsx
@@ -1,7 +1,7 @@
 import React, { useCallback, useEffect, useRef } from "react";
 import type { PaintFile, PaintKeyboardEvent, TextAreaElement } from "paintcannon";
 import { Div, Span, Textarea, useApp } from "paintcannon-react";
-import { useVimKeyHandler, type InputMode } from "./vim-mode.tsx";
+import { DEFAULT_INPUT_MODE, useVimKeyHandler, type InputMode } from "./vim-mode.tsx";
 import { FOREGROUND_COLOR } from "../theme.ts";
 import { ImageInfo } from "../utils/image-utils.ts";
 
@@ -44,7 +44,7 @@ export default function TextInput({
   onImageFilesAttached,
   onRemoveLastImage,
   onSubmit,
-  inputMode = { kind: "emacs" },
+  inputMode = DEFAULT_INPUT_MODE,
   setVimMode,
   onKeyDown,
 }: Props) {
--- a/source/components/text-input.tsx
+++ b/source/components/text-input.tsx
@@ -1,7 +1,7 @@
 import React, { useCallback, useEffect, useRef } from "react";
 import type { PaintFile, PaintKeyboardEvent, TextAreaElement } from "paintcannon";
 import { Div, Span, Textarea, useApp } from "paintcannon-react";
-import { useVimKeyHandler } from "./vim-mode.tsx";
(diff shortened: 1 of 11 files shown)


source/components/text-input.tsx
```diff
@@ -45,14 +44,13 @@ export default function TextInput({
   onImageFilesAttached,
   onRemoveLastImage,
   onSubmit,
-  vimEnabled = false,
-  vimMode = "NORMAL",
+  inputMode = { kind: "emacs" },
```
fireph: I find it a little confusing to have the default inputMode `emacs`. Maybe just a const somewhere:
```
const DEFAULT_INPUT_MODE = { kind: "emacs" };
```
that gets used here is more clear

source/components/vim-mode.tsx
```diff
@@ -3,6 +3,9 @@ import type { CursorVisualPosition, PaintKeyboardEvent, VisualLineRange } from "
 import { useColor } from "../theme.ts";
 import { Span } from "paintcannon-react";
 import { TerminalFlex } from "./terminal-flex.tsx";
+
+export type InputMode = { kind: "emacs" } | { kind: "vim"; mode: "NORMAL" | "INSERT" };
```
fireph: I feel like `type InputMode` should live somewhere other than `vim-mode.tsx`. Maybe `app.tsx`? I'm not sure which file it would best live in.

fireph (approved):
lgtm! Only thing I was thinking about is that this gets rid of most of `vimEnabled` throughout the app in favor of `inputMode`. Does it make sense to also change the config `vimEmulation` to `inputMode`? (e.g. `config.inputMode.kind` = `emacs`/`vim`)

source/components/vim-mode.tsx
sdjidjev: Agreed! Put this in a new file called input-mode.ts!

sdjidjev:
@fireph thanks for the review & comments! 
I agree it makes more sense to change the config's `vimEmulation` field to the more generic `inputMode` but better suited for a followup PR!

sdjidjev:
also @reissbaker PTAL when you can!
i recorded videos of submitted queued messages in vim mode (along with keystrokes) that i put on the pr summary. 
lmk if the behavior looks good to you! also feel free to try out the branch on your machine (or on mine!)

thanks!

reissbaker (commented):
Discussed offline but I think we can skip the Ctrl-Esc shortcut and just assume:

* Esc in insert mode leavesEsc insert mode
* Esc in normal mode interrupts the model

I think that cleans the code up a bit too

PR #285, opened 2026-09-05
Memoize history items

sdjidjev:
Noticed the input field was getting laggy during long context sessions, clanker identified this fix as a quick win!

Wrapped the before/after in a profiler and it showed that render time dropped by about 98%!! (for my long context session)
And typing was way more smooth 😎


fireph (approved):
GoGIF

PR #287, opened 2026-09-11
Starting a process in transport instead of OctoProcess (and clearer tool call UI)

sdjidjev:
- Added clearer tool call UI for `background-process` and `manage-background-process` tool, now showing the actual command it wants to run.
CleanShot 2026-09-10 at 8 17 07 PM@2x

- `manage-background-process` now skips permission requests since it's a safe operation

- Octo will now require a transport in order to start a process. The idea of OctoProcess is no more and now we instead uses `ProcessManager` which keeps track of all running processes and runs a cleanup on any still-running processes (either from `background-process` or otherwise) when Octo exits.


source/transports/transport-process.ts
```diff
@@ -0,0 +1,222 @@
...
+) => void;
+
+export type TransportProcessEvents = {
+  error: [error: Error];
+  spawn: [];
+  exit: [code: number | null, signal: NodeJS.Signals | null];
+  close: [code: number | null, signal: NodeJS.Signals | null];
+};
+
+export interface TransportProcess extends EventEmitter<TransportProcessEvents> {
```
sdjidjev, line 25: this file is mostly a rename of the old `octo-process.ts` but the diff is so large that git won't show them side by side (but mostly everything is the same)

source/transports/shell.ts
```diff
@@ -0,0 +1,123 @@
...
+} from "./transport-common.ts";
+
+const KILL_GRACE_MS = 500;
+
+export function runShell(
+  transport: Transport,
+  signal: AbortSignal,
+  cmd: string,
+  timeout: number,
+  shell = "bash",
```
reissbaker: Generally I'm against default arguments (AI loves them), since it's easy to leave them out and miss something bad during a review. Choosing the default shell, for example, isn't trivial — POSIX only guarantees `/bin/sh` exists, and so for Docker containers we can't assume Bash is installed.

Most local setups have Bash, so we silently use Bash on the local transport since it's a strict superset of `sh` and it's possible an LLM may accidentally use a Bash-ism. But we should be aware during review when and why we're picking a specific shell, since it means Octo will only work on systems that have that shell installed.

source/transports/transport-process.ts
```diff
@@ -0,0 +1,222 @@
...
+import { EventEmitter } from "events";
+import type { Readable, Writable } from "stream";
+import type { TerminateOptions } from "../process-manager.ts";
+
+type ProcessOptions = {
+  surviveAfterOctoExit?: boolean;
+};
+
+export type TransportSpawnOptions = SpawnOptions & ProcessOptions;
+export type TransportExecFileOptions = ExecFileOptions & ProcessOptions;
```
reissbaker: These feel too large to me, and I suspect we don't actually handle all of them correctly especially over Docker (and someday, over SSH). I'd restrict these to just the specific subset of options we want Octo to use.

reissbaker (commented):
Left some comments but generally seems like a reasonable direction to me

sdjidjev:
Closing this to use stacked PRs instead to make it easier to review.

PR #288, opened 2026-09-17
Generic trajectory arcs

reissbaker:
This PR resumes my crusade to migrate Octo's agentic core to a generic `libocto`, which we can eventually publish as a separate package and use on Synthetic and other places.

This PR migrates `trajectory-arc.ts` to a generic `trajectoryArc` function that works on any set of LLM IRs and any set of tools, with type safety. It allows injectable, arbitrary autofix functions, so that `libocto` isn't dependent specifically on our autofix models; any tool can have optional autofix functions, and `state.ts` injects ours for the `edit` tool and for JSON fixing.

I also took the opportunity to revive the core feature request of #259 — automatically retrying failed requests, with request budgeting — since that should live in `trajectory-arc` and I was in here anyway. It works!

image


source/libocto/trajectory-arc.ts
```diff
@@ -0,0 +1,739 @@
...
+        handler.autofixingJson(null);
+        return jsonCorrector(badJson, signal);
+      }
+    : undefined;
+
+  let validationAttempts = 0;
+  let requestAttempts = 0;
+
+  const consumeValidationRetry = (): boolean => {
+    if (validationRetries != null && validationAttempts >= validationRetries) return false;
```
sdjidjev, line 293: i think validationRetries isn't being passed in through `runAgent` right now.
maybe we could make the shape have these retry values as nullable rather than optional

sdjidjev (commented):
i didn't do a full thorough review but i did try the branch out and it works great! feel free to merge~

just one comment about passing in `validationRetries`

source/libocto/trajectory-arc.ts
reissbaker: Oh, that's intentional! Currently in Octo (including prior to this PR) we retry validation failures forever. This change allows callers to specify a retry validation limit since e.g. on Synthetic we wouldn't want to retry forever. A CLI running on your own machine might be different though, where you don't care about limiting those kinds of retries.

The new feature here is: prior to this PR, *request failures* weren't retried at all (e.g. 500s, 400s, etc). This PR also adds request failure retries, although we do backoff + limiting even in Octo there, since it's not just a model behaving weirdly, it's a backend being down.

PR #290, opened 2026-09-17
Move process spawning and cleanup into transports

sdjidjev:
Replace OctoProcess/OctoProcessManager with a TransportProcess wrapper and a ProcessManager cleanup registry.

Transports now own spawn/execFile and background processes, so background processes and LSP servers run on the active transport instead of always on local.

---


source/transports/transport-common.ts
```diff
@@ -209,3 +225,119 @@ export async function getEnvVar(
 ): Promise<string> {
   return (await transport.shell(signal, "echo $" + envVarName, timeout)).replace(/\n$/, "");
 }
+
+const KILL_GRACE_MS = 500;
+
+export function runShell(
```
sdjidjev, line 231: this function was originally `shell()` from `transport/local.ts`.
there was duplication from `runDockerCommand` that was similar to `shell` so i pulled it out into this function in `transport-common.ts`

this should retain the same docker command functionality but now also passes in `-i` which is needed for LSP anyways.

source/transports/transport-process.ts
```diff
@@ -0,0 +1,130 @@
...
+  exit: [code: number | null, signal: NodeJS.Signals | null];
+  close: [code: number | null, signal: NodeJS.Signals | null];
+};
+
+export type TransportProcessOptions = {
+  detached?: boolean;
+  surviveAfterOctoExit?: boolean;
+};
+
+export class TransportProcess extends EventEmitter<TransportProcessEvents> {
```
sdjidjev, line 37: this class/file is mostly just what was in `octo-process.ts` before, but i think there might have been too much renaming plus the file change so git is showing `octo-process.ts` as a full delete, and this new `transport-process.ts` file as a fully new file

source/cli/cli.tsx
```diff
@@ -237,14 +237,9 @@ async function runMain(opts: {
     if (cleanedUp) return;
     cleanedUp = true;
     restoreTitles();
-    await Promise.all([
-      shutdownLspClients(),
-      shutdownMcpClients(),
-      opts.transport.close(),
```
sdjidjev, line 243: `opts.transport.close` is now handled by `terminateOnOctoExit`

source/transports/docker.ts
```diff
@@ -173,32 +120,105 @@ type DockerTarget =
...
+      options,
+    );
+  }
+
+  private commandArgs(
+    command: string,
+    args: readonly string[],
+    options: ProcessSpawnOptions,
+  ): string[] {
+    const dockerArgs = ["exec", "-i"];
```
sdjidjev, line 176: -i was added to make LSPs work because they need an input stream. no other usecases use -i but it's a no-op for them anyways so i left `-i` to be added to all usecases

source/transports/local.ts
```diff
@@ -1,25 +1,79 @@
...
 
 const STRIPPED_ENV_VARS = ["NODE_ENV", "NAPI_RS_NATIVE_LIBRARY_PATH", "CANARY_OCTO"];
 
 export class LocalTransport implements Transport {
   cwd = process.cwd();
-  private readonly octoProcessManager = new OctoProcessManager();
+  // bash over sh: available on most local setups, and tolerant of LLM bash-isms
+  readonly commandShell = "bash";
+  readonly backgroundProcesses: BackgroundProcessManager = new BackgroundProcessManager(this);
+  private readonly runningProcesses = new Set<TransportProcess>();
```
reissbaker, line 24: Nit: since you only ever read from this by turning it into an array with a splat, IMO this should just be an array rather than a Set:

```typescript
private readonly runningProcesses: TransportProcess[] = [];
```

source/background-process.ts
```diff
@@ -110,17 +116,16 @@ export class BackgroundProcessManager {
   private readonly backgroundProcesses = new Map<string, BackgroundProcess>();
   private nextId = 0;
 
-  constructor(private readonly octoProcessManager: OctoProcessManager) {}
+  constructor(private readonly transport: Transport) {}
```
reissbaker, line 119: I think there's a bit too much circular dependency here, where Transports have a BackgroundProcessManager and BackgroundProcessManagers have a Transport. Usually that indicates a bit of a design issue.

I think a simpler approach would be:

1. BackgroundProcessManagers don't have a transport, they just operate on TransportProcesses which are passed in. You don't `backgroundProcessManager.start(command, label)`, you `backgroundProcessManager.track(process, label)`.
2. Transports do not have a BackgroundProcessManager. Instead, we expose a new `backgroundShell` that's similar to the current `shell`, but requires you to pass in a BackgroundProcessManager. The `backgroundProcess` function would spawn the process, and track it in the BackgroundProcessManager.

Callers would then do the following:

```typescript
transport.backgroundShell({
  command: "some-command",
  signal: abortSignal,
  processManager: processes.manager(),
});
```

reissbaker (commented):
Left some nits but looks like a solid direction!

source/transports/local.ts
sdjidjev: i think if we were only adding to this i would agree but the removal is a little more clunky by having to do 
```
void localProcess.processClosedPromise.then(() => {
  this.runningProcesses.splice(this.runningProcesses.indexOf(localProcess), 1);
});
```
vs
```
void localProcess.processClosedPromise.then(() => this.runningProcesses.delete(localProcess));
```

if you think it would be cleaner though happy to change it!

reissbaker: Ohh good point sounds good to keep as a set then

sdjidjev (commented):
Thanks for the review! Addressed comments, PTAL!

reissbaker (approved):
LGTM!

PR #294, opened 2026-09-19
Make menu into modal

sdjidjev:
Turns the menu into a modal that appears over the rest of the app.

<video src="https://github.com/user-attachments/assets/bed8a847-fd59-42a0-94ee-e7ece9e30966" controls="controls" muted="muted" style="max-width: 100%;"></video>

Also allows users to switch models or do other menu commands mid-response!

<video src="https://github.com/user-attachments/assets/6eebeda7-dba2-46d5-9d2e-92814277cd70" controls="controls" muted="muted" style="max-width: 100%;"></video>

Loading a session mid-response aborts the current response and switches to the new session.


reissbaker (approved):
LGTM! Very cool!

Separately, man, the pop-in of the Synthetic Quota display in the header feels even more janky than it used to now that the rest of the menu is so nice. Maybe we should remove it and stick it in a special menu item (`q` for "Quota"?). Feel free to ship this PR as is though and we can do that later!

sdjidjev:
yeah agreed on following up on that! at least the loady pop-in thing only happens on a blank session.. as soon as any conversation happens it's smooth!

